PR #2, opened 2023-03-28
clean up python server tracking doc

tdumitrescu:
Most important bit here is that you should never expose session IDs, any more than you would a password.

### DIFF
--- a/Tracking/how-tos/effective-server.md
+++ b/Tracking/how-tos/effective-server.md
@@ -84,7 +84,7 @@ If _don't_ know the user's identity at the time the event is tracked, then they'
 If you're tracking from servers, you'll need to generate and manage that ID yourself. **Note: If you're doing this, make sure you leave the _user_id_ positional argument empty in your track calls. See the Python code sample below.**
 
 ### Step 1: Generate an ID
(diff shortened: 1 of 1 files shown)


Tracking/how-tos/effective-server.md
```diff
@@ -84,7 +84,7 @@ If _don't_ know the user's identity at the time the event is tracked, then they'
 If you're tracking from servers, you'll need to generate and manage that ID yourself. **Note: If you're doing this, make sure you leave the _user_id_ positional argument empty in your track calls. See the Python code sample below.**
 
 ### Step 1: Generate an ID
-The key is to have an ID that is unique to each user and persists during that user's session. We recommend setting a cookie and using that cookie as the ID. All common server frameworks provide a simple way to set and retrieve cookies per request.
+The key is to have an ID that is unique to each user and persists during that user's session. We recommend generating a random (how many chars??) UUID and storing that value in a cookie. All common server frameworks provide a simple way to set and retrieve cookies per request.
```
tdumitrescu: probably this is fine:
```suggestion
The key is to have an ID that is unique to each user and persists during that user's session. We recommend generating a UUID and storing that value in a cookie. All common server frameworks provide a simple way to set and retrieve cookies per request.
```

ranic (approved):

PR #52, opened 2023-05-04
Update privacy-and-security.md

vicentgilmixpanel:
Fixing link in line 26


ranic:
This link is correct as it is, I think. We do want to point to our developer reference in this case.

PR #83, opened 2023-05-04
Update pricing-plans.md

julianneb-mixpanel:
Add two \ (one to each $)


ranic:
I think this was fixed in https://github.com/mixpanel/docs/pull/91

PR #95, opened 2023-05-04
Update debugging.md

Tofufu:
Updating broken links + added additional github ref links (Changes up to the Data Discrepancy sections).

Notes:
- Events link is broken but can’t fix it because the Events analysis section doesn’t work: https://docs-mixpanel.vercel.app/docs/tracking/how-tos/debugging#debugging-with-events

- Android Debugging needs to be different subsection so I can’t fix the linking there: [Android - Debugging and Logging](https://developer.mixpanel.com/docs/android#debugging-and-logging)


ranic:
Thanks for contributing! I _think_ I was able to address the two outstanding issues.

PR #99, opened 2023-05-04
Update users.md

caracheng:
Added some sections from the missing "Advanced Users Functionality" help doc (https://help.mixpanel.com/hc/en-us/articles/115004695323)

Note - wasn't able to migrate all the content / images over, formatting probably still needs work but just put some work into migrating missing content


ranic:
@caracheng this content is already present here: https://docs-mixpanel.vercel.app/docs/tracking/how-tos/user-profiles -- figured it should live under tracking. 

I do agree we should link to that from the Users doc in the analysis section though. will amend this PR to add a link.

PR #101, opened 2023-05-04
Update what-is-mixpanel.md

emmajaniszewski1:
Formatting fixes.


ranic:
@emmajaniszewski1 fyi I removed the "What is Mixpanel?" heading... by convention in the docs, we don't explicitly put page titles in the page themselves. We can change that convention, but would prefer we do that everywhere rather than just for individual docs.

PR #160, opened 2023-05-23
Event Approval Docs Changes

kavansabnani:
Images aren't rendering for Event Approval doc. I think it's because it's missing. Adding the images back


mavlee:
remove the images under
pages/docs/admin/data-governance/Lexicon-Event-Approval.png

the images should live under
public/

PR #170, opened 2023-05-24
Create rate-limits.md

kathmath:
add API Rate limits


pages/docs/other-bits/rate-limits.md
```diff
@@ -0,0 +1,22 @@
+
```
mavlee, line 1: remove whitespace from top

PR #174, opened 2023-05-24
missing / on lookup tables link

ak--47:
the missing `/` on here causes the page to 404

https://docs.mixpanel.com/docs/tracking/reference/data-model#user-profiles-group-profiles--lookup-tables:~:text=Lookup%20tables%20are%20joined%20to%20events%20and,profile%20properties%20cannot%20be%20lookup%20table%20keys.


jordanmnunez:
Why are you working from a fork of the repo?

PR #207, opened 2023-06-01
Mixpanel customer facing docs for hotshard detection & remediation

amit-saraff:


pages/docs/tracking/reference/distinct-id-limits.md
```diff
@@ -0,0 +1,49 @@
+---
+title: "Distinct ID Limits"
+---
+In order to maintain fast queries and catch implementation mistakes, we set a limit on the number of events sent to a particular `distinct_id` in a given time window. In this document, we explain how this limit works and what to do when you hit it.
+
+## What is a hot shard?
```
entombedvirus, line 6: Should we even expose `shard` as a concept? I think we can explain why sending a large number of events to a single distinct_id and why that's bad without exposing shards as a concept by just saying "hot distinct_id"

pages/docs/tracking/reference/distinct-id-limits.md
```diff
@@ -0,0 +1,49 @@
...
+There are cases when an incorrect implementation results in a disproportionately high number of events sent to Mixpanel for the same `distinct_id`. This leads to shard imbalance where one shard grows larger than the rest impacting storage and query systems which in-turn results in high query latencies for the end user.
+
+## How does hot shard detection work?
+The hotshard detection step runs in the ingestion pipeline. A counter is maintained for each `distinct_id` and `event_date` (derived from `time`) combination and incremented for each event seen with this combination. Once a pre-defined threshold is crossed, the `distinct_id` is marked as a hotshard and all subsequent events for this combination are updated to even the load across shards while keeping all the original event information intact.
```
entombedvirus: > `...threshold is crossed, the `distinct_id` is marked as a hotshard` 

I think that phrase quite doesn't read right. how can a specific distinct_id be a hot shard? I think here we should probably say 

`... threshold is crossed, the distinct_id is marked as a hot distinct_id and all subsequent...`

pages/docs/tracking/reference/distinct-id-limits.md
```diff
@@ -0,0 +1,49 @@
...
+
+The counter is best-effort as a result of the underlying systems used to maintain such a large keyspace.
+
+## What happens when we detect a hot shard?
+Once a given entry crosses the threshold, all subsequent matching events (same `distinct_id` and `event_date`) are updated to a new (hidden) event - `$hotshard_events`. The `distinct_id` is changed to `""` - the system generates its own id resulting in the event ending up in a different (and random) target shard. This prevents one shard from growing disproportionately larger than the others. The original event name and distinct id are stored in the properties object under `mp_original_event_name` and `mp_original_distinct_id`, respectively.
```
entombedvirus: I think this section focuses on the implementation details where I think the more important thing to explain is what and why.

```suggestion
Once a given entry crosses the threshold, all subsequent matching events (same `distinct_id` and `event_date`) will have the following transformations applied to them:

- the event name will be changed to `$hotshard_events`.  The original event name will be preserved under a property called `mp_original_event_name`. Changing the name removes the bad events from being selected for analysis yet remain accessible for debugging.
- `distinct_id` is changed to `""`. The original distinct_id will be preserved under a property called `mp_original_distinct_id`. Removing the distinct_id allows Mixpanel backend to distribute these events evenly ensuring that performance is not adversely affected yet remain accessible for debugging.
```

entombedvirus:
The docs should also provide clarity on:

- does the affected events count toward biling?
- how does identity management factor into hot distinct_id detection? If I have many devices belonging to the same user all sending data does the limit apply to the device_id or the user_id?

pages/docs/tracking/reference/distinct-id-limits.md
amit-saraff: 'shard' as a concept is already defined in public-facing docs - [here](https://docs.mixpanel.com/docs/other-bits/under-the-hood#optimized-for-interactive-user-joins) for example. I am leaning towards using the term 'hotshard' (and adding a line defining what it means based on Kurby's suggestion) in order to tie this terminology across public-facing docs, communication emails, and runbooks.

pages/docs/tracking/reference/distinct-id-limits.md
amit-saraff: I changed wording to specify that 
`the `distinct_id` is marked as contributing to a hot shard`

pages/docs/tracking/reference/distinct-id-limits.md
amit-saraff: That is great callout - updated.

entombedvirus (approved):

PR #240, opened 2023-06-09
Provide guidance on server-side tracking for marketing attribution

chiyc:
Since server SDKs are limited in what they can automatically track, customers will need to implement attribution-related properties manually. We want to help set some expectations and guidance around server-side tracking when it comes to marketing analytics.

### DIFF
--- a/pages/docs/tracking/how-tos/effective-server.md
+++ b/pages/docs/tracking/how-tos/effective-server.md
@@ -27,7 +27,7 @@ mp = mixpanel.init("YOUR_TOKEN")
 
 def track_to_mp(request, event_name, properties):
   parsed = user_agent_parser.Parse(request.headers["User-Agent"])
-  
+
   # Set parsed values as properties.
   # You can also parse out the browser/device/os versions.
   properties.update({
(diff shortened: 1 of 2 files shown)


pages/docs/tracking/how-tos/effective-server.md
```diff
@@ -71,10 +100,18 @@ def handle_signup(request):
...
 ```
 
+## Tracking Page Views
+Page view tracking must be done manually for server-side implementations. Here are some general guidelines for tracking page views.
+
+- Track page views as a single event type by using a constant `event_name`
+- Track different pages as an event property and not as different events for better analysis
+- Fire page view events only on successful responses to the client
+- Parse headers and the request URL for common web analytics properties such as referrer and UTM parameters
+  - See Mixpanel's [default web event properties](/docs/tracking/reference/default-properties#event-properties) for examples
```
ranic: I'd maybe just refer to the other sections of this doc rather than to that default properties page, since that's less helpful than knowing _how_ to track device/browser/location, etc.

pages/docs/tracking/how-tos/effective-server.md
```diff
@@ -71,10 +100,18 @@ def handle_signup(request):
...
   track_to_mp(request, "Signed Up", {"Signup Type": request.form["type"]})
-  
+
   return "Signup successful!"
 ```
 
+## Tracking Page Views
+Page view tracking must be done manually for server-side implementations. Here are some general guidelines for tracking page views.
+
+- Track page views as a single event type by using a constant `event_name`
```
ranic, line 110: We should have some guidance on what to use as the user ID? Maybe refer to the "how to track anonymous users server-side" section of this same doc?

ranic (approved):
Generally LGTM, minor comments

pages/docs/tracking/how-tos/effective-server.md
chiyc: Added the following bullet and sub-bullet on page views:

> - Plan ahead for handling page views from anonymous users, identified users, and connecting them to merge a user's journey
>   - Get started below with [Identifying Users](/docs/tracking/how-tos/effective-server#identifying-users)

pages/docs/tracking/how-tos/effective-server.md
chiyc: Done!

PR #249, opened 2023-06-13
Kurbycchua event user prop

kurbycchua:
Added Reserved Event Properties section and updated Reserved User Properties title and description.


ranic:
Can we add it to this page? https://docs.mixpanel.com/docs/tracking/reference/default-properties

kurbycchua:
@ranic I am ok either way if they stay in the existing structure or moved to https://docs.mixpanel.com/docs/tracking/reference/default-properties The previous reserved user properties was already under the user properties section I mainly updated the title and short blurb. The only thing about putting it under the Default properties is that we may want to split that section to default and reserved

PR #257, opened 2023-06-14
Update formula for weighted average retention rate

mataeuxg:
Update formula in the "When examining a Retention report in % / Retention Rate view" section. C_March 1 and C_March 2 were incorrect.

### DIFF
--- a/pages/docs/analysis/reports/retention.md
+++ b/pages/docs/analysis/reports/retention.md
@@ -247,9 +247,9 @@ From our sample data, the average value for day 1 retained users is calculated a
 
 #### When examining a Retention report in % / Retention Rate view:
 
-C_March 1 = 500 / 1000 (retention rate for March 1)
+C_March 1 = 500 (retention rate for March 1)
 
-C_March 2 = 400 / 500 (retention rate for March 2)
+C_March 2 = 400 (retention rate for March 2)
 
 W_March 1 = 1000 / 1500 (weight for March 1)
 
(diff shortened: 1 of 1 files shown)


pages/docs/analysis/reports/retention.md
```diff
@@ -247,9 +247,9 @@ From our sample data, the average value for day 1 retained users is calculated a
 
 #### When examining a Retention report in % / Retention Rate view:
 
-C_March 1 = 500 / 1000 (retention rate for March 1)
+C_March 1 = 500 (retention rate for March 1)
```
cechoy, line 250: Looks like `C_March 1` refers to the number of retained users in March 1. So the description "retention rate for March 1" is wrong. Should be something like "retained users for March 1" or "count of retained users for March 1".

pages/docs/analysis/reports/retention.md
```diff
@@ -247,9 +247,9 @@ From our sample data, the average value for day 1 retained users is calculated a
...
 
-C_March 2 = 400 / 500 (retention rate for March 2)
+C_March 2 = 400 (retention rate for March 2)
```
cechoy, line 252: Same comment here.

ranic:
Closing since it looks like this was fixed here: https://github.com/mixpanel/docs/pull/271

pages/docs/analysis/reports/retention.md
mataeuxg: Sorry Celia. Just getting back to this. You are correct. Wording should be "count of retained users for March_1".

PR #258, opened 2023-06-14
Fixed snippet when using `MIXPANEL_CUSTOM_LIB_URL` constant

jrbasso:
When using the `MIXPANEL_CUSTOM_LIB_URL` constant as described in https://docs.mixpanel.com/docs/tracking/how-tos/tracking-via-proxy#step-2-point-our-javascript-sdk-at-your-server the extra `n` on the snippet was breaking the implementation.


ranic:
Thanks so much for the contribution!

PR #271, opened 2023-06-21
Update retention.md

dorothy-koh:
Updating value to match sample calculation below


dorothy-koh:
Calculation after this states: 
"**Weighted Average of Retained Users = C_March 1 * W_March 1 + C_March 2 * W_March 2

In our sample data, this computes to: ( 500 * 0.66 ) + ( 400 * 0.33 ) = 466.66"

In this case, C_March 1/2 are both the actual values (500 or 400) rather than 500/1000 or 400/500.

PR #294, opened 2023-06-27
Update events-and-properties.md

amit-saraff:
Add hotshard properties


kurbycchua:
@amit-saraff Just wondering if we should go for a better event Display Name like

mp_original_event_name = Hotshard Original Event Name
mp_original_distinct_id = Hotshard Original Distinct ID

amit-saraff:
@kurbycchua, updated.

PR #305, opened 2023-07-01
Update distinct-id-limits.md

amit-saraff:
Add footnote about numeric distinct id.


kurbycchua:
@amit-saraff looks good thanks for updating

PR #309, opened 2023-07-05
Update tutorials page

naitik-mixpanel:
Serialised it as per mentioned in the index on the left and also added 2 more sections which were missing


naitik-mixpanel:
Closing this since older PR needs to be reverted

PR #318, opened 2023-07-11
Create tips-and-tricks.md

enague:
https://www.notion.so/mxpnl/743859de717547dda7af3b595cca858e?v=7ce8167f857e44228eb0cddcb137848f&p=5cd8115e55494b4c9ddb9152dbf47e24&pm=s


marissakuhrau:
Hey @enague sorry for the delay in reviewing this. New idea – after giving this some more thought, I think we should try using [Github Discussions](https://github.com/orgs/mixpanel/discussions) for weekly Tips & Tricks from Support. It should be easy to incorporate into search results and surface in slack, but the format is a little more friendly for Q&A and it won't become a super long page as we add to it over time. Let's chat more about this but wanted to comment on this for now so it's not hanging out in the open PRs!

PR #358, opened 2023-07-24
Mixpanel/Heliograph branding overrides

eerikson:
What I've done is add a static `mixpanel` theme, which applies an attribute on the root HTML element for us to add style overrides.

This allows us to find the elements/classes we want to override and apply Mixpanel branding to them. I've copied over some of what we're doing on mixpanel.com with theming.

Edit:
Sorry for all the noise & force pushes. This was due to a bug I had with deployments where I had a Node version mismatch (may want to add this to the `engines` field in `package.json` one day).

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,8 @@
     "nextra-theme-docs": "latest",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
-    "sharp": "^0.32.3"
+    "sharp": "^0.32.3",
+    "sass": "^1.64.0"
   },
   "devDependencies": {
     "@types/node": "18.11.10",
--- a/pages/overrides.scss
+++ b/pages/overrides.scss
@@ -1,0 +1,168 @@
+@use "./theme/colors.scss" as color;
+@use "./theme/typography.scss" as typography;
+
+$base-spacing: 8px;
+
+@font-face {
+    font-display: swap;
+    font-family: typography.$garnett;
+    font-style: normal;
+    font-weight: 500;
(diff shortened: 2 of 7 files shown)


package.json
```diff
@@ -25,7 +25,8 @@
     "nextra-theme-docs": "latest",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
-    "sharp": "^0.32.3"
+    "sharp": "^0.32.3",
+    "sass": "^1.64.0"
```
eerikson, line 29: Required for SCSS support

pages/overrides.scss
```diff
@@ -0,0 +1,144 @@
+@use "./theme/colors.scss" as color;
+@use "./theme/typography.scss" as typography;
+
+$base-spacing: 8px;
+
+@font-face {
```
eerikson, line 6: These will go into another import before I merge, bit of visual noise we don't need here

pages/theme/colors.scss
```diff
@@ -0,0 +1,144 @@
...
+// Neutrals
+$black: #000;
+$grey100: #201f24;
+$grey80: #626266;
+$grey50: #9b9b9b;
+$grey30: #c0c0c0;
+$grey10: #e9e9e9;
+$white: #fff;
+
+// Implementations
```
eerikson, line 46: Probably won't need anything below this line, as implementations are different for each app (by definition!)

saminmadani:
Sorry, missed this request. I think @mavlee or @ranic may be better reviewers for this one.

PR #367, opened 2023-07-26
Fix missing tab around Android example section

Coki32:
Android code example section is not in a <Tab></Tab> tag and it is always visible, regardless of tab selection.


ranic:
Thanks for spotting this and fixing it @Coki32! Really appreciate it

PR #396, opened 2023-08-02
Add, Customize UI for Algolia search

eerikson:
Taking the original, non-themed version of Algolia's docsearch plugin:
image

I've looked at where modals like these have been themed across Heliograph, our design system (especially on Pricing). Take a look for yourself, but here's a preview:
image

### DIFF
--- a/components/Search/Search.tsx
+++ b/components/Search/Search.tsx
@@ -1,0 +1,15 @@
+
+import { DocSearch } from '@docsearch/react';
+import '@docsearch/css';
+
+export default function Search() {
+
+
+    return (
+        <DocSearch
+            appId="MQIQQRKVX5"
+            indexName="mixpanel_Docs v2"
+            apiKey="d6267db26ac89477a9a87ea82da493b7"
+            insights={true} />
+        )
+}
--- a/pages/overrides.scss
+++ b/pages/overrides.scss
@@ -2,6 +2,9 @@
 @use "./theme/typography.scss" as typography;
 
 $base-spacing: 8px;
+$border-radius: 8px;
+$border-radius-md: 24px;
+$border-radius-lg: 100px;
 
 @font-face {
     font-display: swap;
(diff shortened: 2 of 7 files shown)


components/Search/Search.tsx
```diff
@@ -0,0 +1,15 @@
...
+import '@docsearch/css';
+
+export default function Search() {
+
+
+    return (
+        <DocSearch
+            appId="MQIQQRKVX5"
+            indexName="mixpanel_Docs v2"
+            apiKey="d6267db26ac89477a9a87ea82da493b7"
```
eerikson, line 12: Should probably put this in an env configuration, but I'm not too worried about exposing this information right now.

components/Search/Search.tsx
```diff
@@ -0,0 +1,15 @@
+
+import { DocSearch } from '@docsearch/react';
+import '@docsearch/css';
+
+export default function Search() {
+
+
+    return (
+        <DocSearch
+            appId="MQIQQRKVX5"
```
tiffanyqi, line 10: is it ok to keep this appid here?

pages/overrides.scss
```diff
@@ -164,5 +167,114 @@ $base-spacing: 8px;
         font-weight: 500;
     }
 
+    // Custom Search theming from Algolia Docsearch
+    .DocSearch {
+        --docsearch-primary-color: #{color.$purple200};
```
tiffanyqi, line 172: do we need to use a hash because we're using these -- props?

pages/overrides.scss
```diff
@@ -164,5 +167,114 @@ $base-spacing: 8px;
...
+        --docsearch-highlight-color:  #{color.$purple100};
+        --docsearch-spacing: #{$base-spacing * 3};
+        --docsearch-modal-width: 47rem;
+        --docsearch-container-background: rgba(27,11,59,.85);
+        --docsearch-modal-background: white;
```
tiffanyqi, line 177: probably use color.white?

tiffanyqi (approved):
does this custom ui override the existing search bar? is it supposed to open the modal when you click on search?

code lgtm

pages/overrides.scss
eerikson: Exactly. Not sure why, but without these SCSS will just spit out the literal variable names for CSS vars.

components/Search/Search.tsx
eerikson: ~~Should be fine if we don't expose it with the secret!~~

Edit: I'm keeping both. I was having a lot of trouble getting the env settings to work on this client-mounted component, and both are exposed via XHR anyway.

eerikson:
> does this custom ui override the existing search bar?

yeah, it renders an empty div instead of the existing Nextra search element, and then puts itself into that.

> is it supposed to open the modal when you click on search?

Correct—bit of an odd UI choice, but I don't think we have a way to override it.

eerikson:
I've decided to remove the env variables as I couldn't find a way to get them working in this app properly, as virtually all of it is client-side rendered. Given that all the above details are exposed in XHR anyway, it didn't seem like a big security issue.

PR #430, opened 2023-08-10
Create data-warehouse

kavansabnani:
Adding a top level page for Warehouse Connectors under Tracking below Server

@ishamehramixpanel @ranic - can you help move the BigQuery and Snowflake pages from under Integrations to under this new Data Warehouse page? Also can we find a way to capitalize the Q in BigQuery


ranic:
@kavansabnani two things:
* Can you add a `.md` suffix to this file?
* Can you add a directory called docs/tracking/data-warehouse (same name as this markdown file, just without the extension) and move the bigquery and snowflake docs inside of that? That will do the nesting

PR #432, opened 2023-08-11
Template help doc update

jaykimes11:


argenisf:
@jaykimes11 @ishamehramixpanel (mentioning both since both have commits), any reason why we go into details of the value proposition for 3 of the board templates but we don't mention the attribution or ecommerce ones? If we go into detail on each one, looks off to not also include those 2, no?

jaykimes11:
@argenisf We have an updated structure coming soon which will cover ecommerce and marketing templates but this needs to be pushed in the meantime.

PR #466, opened 2023-08-23
Update mixpanel-bigquery-export-design.md

dneralla:


haidang-dam (approved):

saminmadani:
I do not think using a codeblock here for this warning is the best design scheme. @tarao-mixpanel any ideas on how to display a warning in the docs? I couldn't find any past examples in the existing docs.

ranic:
For now lets just say: "Note: ..." with no formatting. We can revisit all the "Note:" text later, once we have a callout component.

ishamehramixpanel (approved):

PR #469, opened 2023-08-29
Create 2023-08-29-boards-on-boards.mdx

tiffany-mixpanel:
Changelog for boards on boards - can you please make sure the image uploaded correctly? I guessed a little bit here


mherrman (commented):
Hey Tiffany I think the image needs to be uploaded to the public/changelog folder and then have the path changed to:

/changelog/Boards_on_Boards.gif

Right now the image is in the /public i think and the examples for changelog images specifically i've seen in reviewing all are in that changelog folder

mherrman (approved):

PR #474, opened 2023-08-30
Update migrating-to-mixpanel-from-amplitude.md

PeishanFromMixpanel:
Added note to ensure that projects are using simplified id merge if updating amp sdk to ours. 

Edited note on who to contact - success@mixpanel.com vs. support


argenisf (approved):
LGTM

PR #501, opened 2023-09-06
Update export.openapi.yaml

Tofufu:
Removing "The return is format is one event per line sorted by increasing timestamp." for Raw Export API.  There currently isn't sorting with Export API (see SRFE-4689).


Tofufu:
Closing out pull request - looks like there was another PR addressing this that is already merged: https://github.com/mixpanel/docs/pull/503

PR #521, opened 2023-09-13
Update user-profiles.md

emmajaniszewski1:
Some text is inadvertently being rendered as LaTeX formulas due to the integration used elsewhere in the documentation. See screenshot for an example of the bug in question.

The $ needs to be escaped for it not to mark the beginning of a LaTeX formula.

Screen Shot 2023-09-13 at 2 18 06 PM


argenisf (approved):
LGTM

PR #538, opened 2023-09-20
added guide on how to rotate API secret

leonardofratini:


ranic (approved):
Nit: You can use ⚙️ instead of Cogwheel Icon

PR #548, opened 2023-09-26
Nav restructuring

tiffanyqi:
plus styling
Screen Shot 2023-09-29 at 11 23 36 AM


pages/overrides.scss
```diff
@@ -142,14 +143,70 @@ $border-radius-lg: 100px;
...
+
+        // remove the vertical rule within dropdowns
+        .before\:nx-w-px:before {
+            width: 0;
+        }
+
+        .hover\:nx-bg-gray-100:hover {
+            background-color: colors.$purple20;
+            color: colors.$purple140;
+        }
```
tiffanyqi: forgot to comment this, but this was to modify the hover state

pages/overrides.scss
```diff
@@ -142,14 +143,70 @@ $border-radius-lg: 100px;
...
+        li {
+            margin-left: 4px;
+            margin-right: 4px;
+        }
+
+        .nx-pt-1 {
+            padding-bottom: 0.25rem;
+        }
+    }
+
```
tiffanyqi: these modified the focus states

pages/overrides.scss
```diff
@@ -142,14 +143,70 @@ $border-radius-lg: 100px;
...
+            color: colors.$purple140;
+        }
+
+        // modify the non-focused nav items
+        .nx-text-gray-500 {
+            color: colors.$purple200;
+        }
+
+        // bold the parent links
+        .nextra-scrollbar >.nx-transform-gpu > .nx-transition-opacity > .nx-flex > li > button,
```
eerikson: I like what you've had here with these immediate descendant selectors—they at least give us some idea in the future about what the intent was if something breaks.

Is there any equivalent we can do above and below, for example on line 192 and  153?

theme.config.tsx
```diff
@@ -52,7 +53,38 @@ const config: DocsThemeConfig = {
   },
   primaryHue: { dark: 256, light: 259 },
   sidebar: {
+    defaultMenuCollapseLevel: 1,
     toggleButton: true,
+    titleComponent: ({ title, type }) => {
+      if (type === `separator`) {
+        let icon;
+        switch (title) {
+          case `INTRO`:
```
eerikson, line 62: Is there any way we can infer these titles from elsewhere? Not a big deal if not.

eerikson (approved):
This is WILD, thanks so much for taking care of this. Just left a few quesitons/nits that I think shouldn't block you.

theme.config.tsx
tiffanyqi: they come from a `_meta.json` file... but I'm not sure we can share variables between the two

pages/overrides.scss
tiffanyqi: will try!

tiffanyqi:
Closing in favor of https://github.com/mixpanel/docs/pull/582

PR #602, opened 2023-10-10
Stage reference changes to readme

tiffanyqi:
This github workflow:

1) creates a new readme version, if it needs to (i.e. multiple pushes on the pr does not generate multiple readme versions)
2) pushes content from the docs/reference folder onto that staging link
3) adds/updates a comment on the PR and links to that staging link, and then finally
4) upon merge, deletes the readme version.


.github/workflows/rdme-delete-staging.yml
```diff
@@ -0,0 +1,15 @@
+name: Delete ReadMe Staging
```
tiffanyqi, line 1: workflow untested. how does this look? does it have to be in a new file, or can we combine it with the below? i don't want this to run at all if the pr isn't merged, and i also don't want it firing an alert saying that nothing was run if we don't have to

.github/workflows/rdme-staging.yml
```diff
@@ -0,0 +1,89 @@
...
+    outputs:
+      has-version: ${{ steps.check-readme-version.outputs.has-version }}
+    steps:
+      - name: Check out repo 📚
+        uses: actions/checkout@v3
+
+      # installs readme so that we can run `rdme` in the conditional step
+      - name: Install rdme
+        run: |
+          npm install -g rdme
```
tiffanyqi: is this the right command? is it ok to install rdme globally each time this workflow runs?

.github/workflows/rdme-staging.yml
```diff
@@ -0,0 +1,89 @@
...
+
+      - name: Check if the readme version exists
+        id: check-readme-version
+        run: |
+          if rdme versions --version=${{ github.event.number }} --key=${{ secrets.README_API_KEY }}; then
+            echo "has-version=true" >> $GITHUB_ENV
+          else
+            echo "has-version=false" >> $GITHUB_ENV
+          fi
```
tiffanyqi: is this the right way to check whether the version already exists? tbh, i got this from chat gpt, so i'm unsure if this is up to date / too much!

.github/workflows/rdme-staging.yml
```diff
@@ -0,0 +1,89 @@
...
+        with:
+          rdme: versions:create ${{ github.event.number }} --key=${{ secrets.README_API_KEY }} --fork=v3.26 --main=false --beta=false --deprecated=false --isPublic=true
+
+      - name: Push docs to staging env 🚀
+        uses: readmeio/rdme@v8
+        with:
+          rdme: docs ./reference --key=${{ secrets.README_API_KEY }} --version=${{ github.event.number }}
+
+  # build and update comment with the proper link
+  preview-notify:
```
tiffanyqi, line 45: took this whole section from the marketing-site yaml. the only thing I modified here was in `sparse-checkout` and the api reference preview link.

tiffanyqi:
@austinpray-mixpanel how do I remove the `push` workflows on both rdme-staging and preview / notify? i think the only way to get the pull request numbers is to set them as such in `on`, but when i move the branches-ignore to `pull_request`, the workflow wouldn't run at all

.github/workflows/rdme-staging.yml
```diff
@@ -0,0 +1,89 @@
+name: Generate ReadMe Staging 🦉
+
+on:
+  # when a PR is created in any branch except main that have made changes to the
+  # reference folder, run this workflow
+  push:
+    branches-ignore:
+      - main
```
austinpray-mixpanel:
```suggestion
```

you can omit this entirely and it will run on all pull_request updates

.github/workflows/rdme-staging.yml
```diff
@@ -0,0 +1,89 @@
...
+jobs:
+  # ////////////////////////////////////////////////////////////////////////
+  # Pull Request
+  # ////////////////////////////////////////////////////////////////////////
+
+  # Create a new ReadMe Version if it needs to and pushes content to the version
+  rdme-staging:
+    runs-on: ubuntu-latest
+    outputs:
+      has-version: ${{ steps.check-readme-version.outputs.has-version }}
```
austinpray-mixpanel: what is this for?

.github/workflows/rdme-staging.yml
```diff
@@ -0,0 +1,89 @@
...
+
+      # installs readme so that we can run `rdme` in the conditional step
+      - name: Install rdme
+        run: |
+          npm install -g rdme
+
+      - name: Check if the readme version exists
+        id: check-readme-version
+        run: |
+          if rdme versions --version=${{ github.event.number }} --key=${{ secrets.README_API_KEY }}; then
```
austinpray-mixpanel: I expected something like `--version=pr${{ github.event.number }`

.github/workflows/rdme-staging.yml
austinpray-mixpanel: Do it like this https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#example-of-setting-an-output-parameter

use $GITHUB_OUTPUT instead of GITHUB_ENV

.github/workflows/rdme-staging.yml
```diff
@@ -0,0 +1,89 @@
...
+      - name: Run create readme version
+        if: steps.has-version.outputs.has-version == 'false'
+        uses: readmeio/rdme@v8
+        with:
+          rdme: versions:create ${{ github.event.number }} --key=${{ secrets.README_API_KEY }} --fork=v3.26 --main=false --beta=false --deprecated=false --isPublic=true
+
+      - name: Push docs to staging env 🚀
+        uses: readmeio/rdme@v8
+        with:
+          rdme: docs ./reference --key=${{ secrets.README_API_KEY }} --version=${{ github.event.number }}
```
austinpray-mixpanel: I expected `--version=pr${{ github.event.num }}` so the version number is like vpr123 instead of v123

.github/workflows/rdme-staging.yml
austinpray-mixpanel: yup looks good! sparse checkout is less relevant for tiny docs repo but can't hurt!

austinpray-mixpanel (changes requested):

.github/workflows/rdme-staging.yml
tiffanyqi: i don't know, i left this here because that's what chat gpt contained lol. ultimately, i need to check to see if readme has already created a new version, so that's what the `if` checks were for on line 35. do i need the above line in conjunction with that line 35 block?

.github/workflows/rdme-staging.yml
tiffanyqi: sure! i can modify that.

.github/workflows/rdme-staging.yml
tiffanyqi: like this?
```
        run: |
          if rdme versions --version=pr${{ github.event.number }} --key=${{ secrets.README_API_KEY }}; then
            echo "has-version=true" >> "$GITHUB_OUTPUT"
          else
            echo "has-version=false" >> "$GITHUB_OUTPUT"
          fi

```

what's the difference here? would I have to modify anything else to use this?

.github/workflows/rdme-staging.yml
austinpray-mixpanel: Then it's not needed. You can just do like the example here: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#example-of-setting-an-output-parameter

.github/workflows/rdme-staging.yml
```diff
@@ -0,0 +1,83 @@
...
+        run: |
+          if rdme versions --version=pr${{ github.event.number }} --key=${{ secrets.README_API_KEY }}; then
+            echo "has-version=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "has-version=false" >> "$GITHUB_OUTPUT"
+          fi
+
+      # if it's a new PR, we'll have to create the readme, otherwise we can skip this step
+      - name: Run create readme version
+        if: steps.has-version.outputs.has-version == 'false'
```
austinpray-mixpanel:
```suggestion
        if: steps.check-readme-version.outputs.has-version == 'false'
```

.github/workflows/rdme-staging.yml
austinpray-mixpanel: yeah it's okay to install globally but I would prolly pin the version in a package.json

tiffanyqi: our `rdme` package is pinned in package.json to 8.6.6, is that enough in this case here?

.github/workflows/rdme-staging.yml
```diff
@@ -0,0 +1,76 @@
...
+          if rdme versions:create 0.0-pr-${{ github.event.number }} --key=${{ secrets.README_API_KEY }} --fork=v3.26 --main=false --beta=false --deprecated=false --isPublic=true; then
+            echo "Version created successfully"
+          else
+            echo "Errors because the version already exists, and so we can continue to the next step"
+          fi
+
+      - name: Push docs to staging env 🚀
+        uses: readmeio/rdme@v8
+        with:
+          rdme: docs ./reference --key=${{ secrets.README_API_KEY }} --version=0.0-pr-${{ github.event.number }}
```
tiffanyqi, line 42: sorry for all of the commits / pushes! this is the only thing that changed from last you looked

.github/workflows/rdme-delete-staging.yml
austinpray-mixpanel: This looks like it should work fine. If it doesn't work or is too hard to develop we can just turn this into a cron that looks for versions like 0.0-pr-$PR_NUMBER and then for every $PR_NUMBER check if it's merged and delete if so.

.github/workflows/rdme-staging.yml
austinpray-mixpanel: PRed your PR here https://github.com/mixpanel/docs/pull/608

.github/workflows/rdme-delete-staging.yml
```diff
@@ -0,0 +1,15 @@
...
+      paths:
+      - 'reference/**'
+
+jobs:
+  delete-staging:
+    if: github.event.pull_request.merged == true
+    runs-on: ubuntu-latest
+    uses: readmeio/rdme@v8
+    with:
+      rdme: versions:delete 0.0-pr-${{ github.event.number }} --key=${{ secrets.README_API_KEY }}
```
austinpray-mixpanel: Oh should probably check that the version exists before deleting huh?

You can wrap this in a little bash script so you can test outside of the action

austinpray-mixpanel (changes requested):
PRed your PR https://github.com/mixpanel/docs/pull/608

.github/workflows/rdme-delete-staging.yml
```diff
@@ -0,0 +1,27 @@
+name: Delete ReadMe Staging
+on:
+  pull_request:
+    types:
+      - closed
+      paths:
+      - 'reference/**'
```
austinpray-mixpanel:
```suggestion
```

This looks like wrong indention. And I would just run this always anway.

tiffanyqi: but wouldn't it be better to do so only a PRs that created the staging env in the first place? (will def fix the indention, was going to push it!)

austinpray-mixpanel: Yeah it would be, would need to match the parent workflow's paths for that to not leak versions

image

austinpray-mixpanel (approved):
lgtm!

PR #606, opened 2023-10-11
Fix replace lookup table api

haidang-dam:
While implementing lookup table import on warehouse, I realized this error in docs while trying to debug.


mavlee (approved):

mavlee:
hmm actually this is faiing checks, but not sure why offhand

PR #618, opened 2023-10-16
Add the description property to Lexicon's events

vfragosop:
I've noticed the description is missing from `results[].jsonSchema.description`, which causes generated SDKs to also miss the property.


ranic (approved):

vfragosop:
I've rebased this branch onto `main` to account for the removal of `openapi/out` on #648.

PR #630, opened 2023-10-22
Dynamic token in JS init snippet

santigracia:
adding dynamic token for the JS init example snippet using url parameter when present

### DIFF
--- a/pages/docs/tracking/javascript-quickstart.mdx
+++ b/pages/docs/tracking/javascript-quickstart.mdx
@@ -30,11 +30,39 @@ yarn add mixpanel-browser
 ## Step 2: Track your first event
 
 You'll need your Project Token for this step, which you can get [here](https://mixpanel.com/settings/project).
+
+import { useRef } from 'react'
+import { useRouter } from 'next/router';
+
+export function DynamicCode({ children }) {
+  const ref = useRef()
+  const tokenRef = useRef()
+  const router = useRouter();
+  const valueFromURL = router.query.projectToken;
+
+  if (ref.current) {
+    if (valueFromURL) {
+      const token = [...ref.current.querySelectorAll('code span')].find(
+        el => el.innerText === '\'YOUR_TOKEN\''
+      )
+      token.innerText = '\'' + valueFromURL + '\''
+    }
+  }
+  return (
+    <>
+      <div ref={ref}>
+        {children}
+      </div>
+    </>
+  )
+}
+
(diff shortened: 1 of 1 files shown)


pages/docs/tracking/javascript-quickstart.mdx
```diff
@@ -30,11 +30,39 @@ yarn add mixpanel-browser
 ## Step 2: Track your first event
 
 You'll need your Project Token for this step, which you can get [here](https://mixpanel.com/settings/project).
+
+import { useRef } from 'react'
+import { useRouter } from 'next/router';
+
+export function DynamicCode({ children }) {
+  const ref = useRef()
+  const tokenRef = useRef()
```
tiffanyqi, line 39: we're not using this, can remove

pages/docs/tracking/javascript-quickstart.mdx
```diff
@@ -30,11 +30,39 @@ yarn add mixpanel-browser
...
+  const router = useRouter();
+  const valueFromURL = router.query.projectToken;
```
tiffanyqi, line 41: wow i didn't know about this, this is so cool!

tiffanyqi (commented):
this works great, thanks for your hard work Santi! A couple of things:
1. We are actually going to remove this file soon. Any chance we can hold off on the deployment until we release the connect-your-data file? Should go out today, looks like this: https://github.com/mixpanel/docs/pull/582/files#diff-41c87c30a0527c677d0ead4ea37f98986cd01bb671964d67c77d8b7b23075602

2. The use of the project token is actually in quite a few places, not just the javascript quickstart. With the above new article being created, we'll need it in every SDK. Can we make it so that this file is more re-usable / applied to each section? it could just be us making it `YOUR_TOKEN` or something everywhere, but wanted to double check in case you had any other ideas. Thanks!

pages/docs/tracking/javascript-quickstart.mdx
```diff
@@ -30,11 +30,39 @@ yarn add mixpanel-browser
...
+
+  if (ref.current) {
+    if (valueFromURL) {
+      const token = [...ref.current.querySelectorAll('code span')].find(
```
tiffanyqi, line 45: i'd like to avoid the use of querySelectorAll if possible, but if not that's ok!

myronkaifung:
Hi @santigracia , would you please review Tiffany's comments?

PR #634, opened 2023-10-23
Update reference to sso-support

RyanBuccellato:
Should go to regular support now, updated references in the code but forgot to make a PR for the docs too.

### DIFF
--- a/pages/docs/admin/sso.md
+++ b/pages/docs/admin/sso.md
@@ -78,7 +78,7 @@ There are [instructions here to set up SSO with Azure.](/docs/admin/sso/azure) A
 
 **Other IDPs**
 
-It is possible to set up Mixpanel with IDPs not listed above. Email sso-support@mixpanel.com to get the required information to set up SSO with an IDP not listed.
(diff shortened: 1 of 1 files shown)


pages/docs/admin/sso.md
```diff
@@ -78,7 +78,7 @@ There are [instructions here to set up SSO with Azure.](/docs/admin/sso/azure) A
 
 **Other IDPs**
 
-It is possible to set up Mixpanel with IDPs not listed above. Email sso-support@mixpanel.com to get the required information to set up SSO with an IDP not listed.
+It is possible to set up Mixpanel SSO with IDPs not listed above. [Contact support](https://mixpanel.com/get-support) for further assistance in such cases.
```
RyanBuccellato, line 81: feel free to change the wording, just changed it to be a bit more generic since we may not always have a specific set of instructions, tends to be more hands on

mherrman (approved):

pages/docs/admin/sso.md
mherrman: haha thanks Bucc! Working even when you just left. We are gonna miss you over here on support big time. Enjoy the time off!

PR #660, opened 2023-10-27
JSON Properties Column Docs

jakewski:
Added some documentation around choosing a JSON properties column. Some additional changes:

- Made column mappings in event a table to scan more easily
- Added a small page for "Column Types" to detail what types of columns we recognize per warehouse
- Added "sending lookup tables" to _meta.json so that it appears in order on the nav


aswani521 (approved):
lgtm

aswani521 (approved):
lgtm

PR #674, opened 2023-11-01
Add openapi to delete staging paths

tiffanyqi:
noticed that delete-staging wasn't running on PRs that included openapi/, so now it's here

### DIFF
--- a/.github/workflows/rdme-delete-staging.yml
+++ b/.github/workflows/rdme-delete-staging.yml
@@ -5,6 +5,7 @@ on:
       - closed
     paths:
       - 'reference/**'
+      - 'openapi/**'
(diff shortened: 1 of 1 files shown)


junhouse (approved):
So not removing `readme` was stopping us from removing staging-links for the PRs that had changes in `openapi` folder?

tiffanyqi:
Chatted with Jun offline, but essentially the context of this PR is that we create new staging links (READMe versions) whenever a change to the openapi/ folder is created. When the PR is merged or closed, we run this workflow to delete the version from READMe. openapi/ wasn't added to this workflow, but now it is, so versions associated with those PRs should now be cleaned up.

PR #700, opened 2023-11-15
Add thumbs up/down component to docs

tiffanyqi:
Screen Shot 2023-11-14 at 4 44 45 PM

Hover:
Screen Shot 2023-11-14 at 4 44 54 PM

Focus:
Screen Shot 2023-11-14 at 4 45 12 PM

Response:
Screen Shot 2023-11-14 at 4 57 56 PM


components/FeedbackCollector/FeedbackCollector.module.scss
```diff
@@ -0,0 +1,35 @@
+.buttonContainer {
+    display: flex;
+    flex-direction: row;
+}
+
+.button {
+    border: 1px solid #c0c0c0;
```
junhouse: So in `docs` we are more lenient re: the code structure?

components/FeedbackCollector/FeedbackCollector.tsx
```diff
@@ -0,0 +1,46 @@
...
+import ThumbsUpIcon from "../svg/ThumbsUp";
+
+import style from "./FeedbackCollector.module.scss";
+
+export function FeedbackCollector() {
+  const [gaveFeedback, setGaveFeedback] = useState(false);
+
+  const handleFeedback = function (value: boolean) {
+    const props = {
+      title: document.querySelector(`h1`).textContent,
```
junhouse, line 15: yea.. there does not seem like a good way to grab the title. Do you think it's more reliable to grab the `url path` here? so for example,
`https://docs.mixpanel.com/docs/how-it-works/infrastructure` -> we could grab the last `path`?

components/FeedbackCollector/FeedbackCollector.tsx
```diff
@@ -0,0 +1,46 @@
...
+import { track } from "../../utils/tracking";
+import ThumbsDownIcon from "../svg/ThumbsDown";
+import ThumbsUpIcon from "../svg/ThumbsUp";
+
+import style from "./FeedbackCollector.module.scss";
+
+export function FeedbackCollector() {
+  const [gaveFeedback, setGaveFeedback] = useState(false);
+
+  const handleFeedback = function (value: boolean) {
```
junhouse: nit (not blocking): `value` -> `isPositiveFeedback` ?

components/FeedbackCollector/FeedbackCollector.tsx
junhouse: Also, i think it would be nice to capture the full url path too, so we know which page got thumbs up OR thumbs down.

junhouse (approved):
LGTM, I left a suggestion re: grabbing the title / capturing the url path.

components/FeedbackCollector/FeedbackCollector.module.scss
tiffanyqi: very much so. however, you've reminded me that we have a colors file, so i'll import the grey and lava at least.

components/FeedbackCollector/FeedbackCollector.tsx
tiffanyqi: we already have the url path, as it's sent to mixpanel by default.

if we just use the url path for the title, it's pretty much the same thing, right? plus, spaces will have dashes in it. not sure, i don't want to overcomplicate it

junhouse (approved):

PR #701, opened 2023-11-15
Vchenmoney patch 1

vchenmoney:
- revised data model differences
- added instrumentation diagram
- revised adobe key metrics


saminmadani (approved):
lgtm, will defer to you on accuracy and terminology for Adobe :)

PR #710, opened 2023-11-16
fix google ads example code.

ak--47:
the node.js SDK requires an API secret when using the import_batch() method: https://github.com/mixpanel/mixpanel-node/blob/d304d857553ee786984e3b764f3d2140196d2b62/lib/mixpanel-node.js#L116

the code example we provide for google ad spend data has the secret at the top of the file, but doesn't properly use the secret on the `init()` ... this results in an error when `import_batch()` gets called, preventing our example code from working as is.

this change fixes the `init()` step so the code in our guide doesn't  require any additional debugging.


saminmadani (approved):
lgtm, feel free to merge whenever

PR #739, opened 2023-11-30
hydration error on safari mobile

tiffanyqi:
trying to fix https://app.rollbar.com/a/mixpanel/fix/item/interactive-web/716#occurrences
issue: https://nextjs.org/docs/messages/react-hydration-error

I've been unable to test it on my machine, but I can repro it on my phone. My guess is that there are some links that do not appear the same in production, and so I've tried to convert all `a` into native links and set emails the same.

affects:
- https://docs.mixpanel.com/docs/tracking-methods/sdks/nodejs
- https://docs.mixpanel.com/docs/features/custom-properties
- https://docs.mixpanel.com/docs/tracking-methods/sdks/javascript
- https://docs.mixpanel.com/docs/privacy/protecting-user-data# -> not sure how to fix this one


eerikson (approved):
Weird... let me know if you could use another pair of eyes on this.

PR #740, opened 2023-12-01
Specify custom pageview tracking; add custom attribution

eerikson:
This synchronizes some custom pageview tracking we're using back in the marketing site.

In short, the SDK will track these parameters at a session level; for example, `utm_campaign=foobar` will persist across visits for each user. We don't want to change how the SDK does this for everybody, but we still want to change this to work _per page_ only, so that we have more flexibility at the analysis layer.

So, now, when `utm_campaign=foobar` is present in the current page URL (and only in the current page URL), we will track it as a custom property like so:

```JavaScript
track_pageview({
    [`page.utm_campaign`]: `foobar`,
})
```

Notice that it is namespace with `page.` so that we don't overwrite or collide with how the SDK tracks attribution currently.


components/GTMScripts/GTMScripts.tsx
```diff
@@ -36,7 +36,7 @@ export default function insertGTMScriptTags() {
             api_payload_format: 'json',
             debug: ${!isProd},
             persistence: 'localStorage',
-            track_pageview: true,
+            track_pageview: false,
```
eerikson, line 39: Disable automatic pageview tracking. Unfortunately, this doesn't support any custom properties we may want to attach.

package.json
```diff
@@ -36,6 +36,7 @@
   },
   "devDependencies": {
     "@redocly/openapi-cli": "1.0.0-beta.16",
+    "@types/mixpanel": "^2.14.8",
```
eerikson, line 39: Added types for the Mixpanel SDK. Even though we have to extend them slightly, it's good to have!

types.d.ts
```diff
@@ -0,0 +1,11 @@
+
+interface Mixpanel {
+  // @TODO: https://mixpanel.atlassian.net/browse/INTRACTIVE-685
```
eerikson, line 3: These refer to a ticket where we should contribute these back to the community at some point.

pages/_app.tsx
```diff
@@ -4,12 +4,93 @@ import React from "react";
...
+  mixpanel.track_pageview(namespacedAttribution);
+}
+
 
 export default function MyApp({ Component, pageProps }: AppProps) {
+  const router = useRouter();
+
   useEffect(() => {
     insertGTMScriptTags();
+    trackPageview();
```
eerikson, line 81: This will track on the initial page load.

pages/_app.tsx
```diff
@@ -4,12 +4,93 @@ import React from "react";
...
     // @ts-ignore
   }, []);
 
-  return <Component {...pageProps} />;
+  useEffect(() => {
+    router.events.on(`routeChangeComplete`, trackPageview);
```
eerikson, line 86: This will track on subsequent page transitions.

package-lock.json
```diff
@@ -13,7 +13,7 @@
         "@docsearch/react": "^3.5.2",
         "clsx": "^1.2.1",
         "next": "^13.0.6",
-        "nextra": "*",
+        "nextra": "latest",
```
eerikson, line 16: Are these the same thing? Is it safe to do either of these without pinning?

pages/_app.tsx
```diff
@@ -4,12 +4,93 @@ import React from "react";
...
+  const namespacedAttribution = cloneWithoutFalseyProperties({
+      [`page.gclid`]: pageviewAttribution.gclid,
+      [`page.utm_source`]: pageviewAttribution.utmSource,
+      [`page.utm_term`]: pageviewAttribution.utmTerm,
+      [`page.utm_campaign`]: pageviewAttribution.utmCampaign,
+      [`page.utm_content`]: pageviewAttribution.utmContent,
+      [`page.utm_medium`]: pageviewAttribution.utmMedium,
+      [`page.partner_key`]: pageviewAttribution.partnerKey,
+  });
+  mixpanel.track_pageview(namespacedAttribution);
```
junhouse, line 72: How were we tracking these attributes before? does setting `track_pageview = true` automatically captures these attributes?

pages/_app.tsx
```diff
@@ -4,12 +4,93 @@ import React from "react";
...
+  // Filter for any null values
+  const cleanedParams = cloneWithoutFalseyProperties(obtainedParams);
+  return cleanedParams;
+}
+
+
+function trackPageview() {
+  const pageviewAttribution = extractParamsFromBrowser();
+  const namespacedAttribution = cloneWithoutFalseyProperties({
+      [`page.gclid`]: pageviewAttribution.gclid,
```
junhouse, line 64: So, does calling `.track_pageview` fires off the same event as we did when we had (`track_pageview = true`)? If that's the case, what's the issue with collision? Not sure why we need `page.` prefix here if they are all the same events.

junhouse (commented):
a few qs for my understanding.

pages/_app.tsx
eerikson: From my understanding, `track_pageview  = true` basically means that `track_pageview ()` will immediately be called. Because we want to bundle custom event properties as an argument, we'll have to call this manually.

`track_pageview ()` _does_ capture these attributes but in a totally different method: session-based. It will persist across pages and visits. We don't want to do that here.

junhouse: Ah, understood - those are captured in the server (per your previous PR + Chenxing's change).

pages/_app.tsx
eerikson: Because these are different property values:
```
utm_campaign: "foobar",
page.utm_campaign: "foobar"
```

`utm_campaign` is a default property that mixpanel tracks. However, there is logic in the SDK that make the value different depending on browser state. For example, take this user flow:

1. User lands on `page.com/?utm_campaign=foobar`:
    ```
        utm_campaign: "foobar",
        page.utm_campaign: "foobar"
    ```

2. User navigates to `page.com/baz`:
    ```
        utm_campaign: "foobar",
        page.utm_campaign: undefined
    ```
3. User closes tab & leaves `page.com`.
4. The next day, the user comes back to `page.com`:
    ```
        utm_campaign: "foobar",
        page.utm_campaign: undefined
    ```


This is because the SDK has a custom definition of how to persist `utm_campaign` that we don't want to use. The reason why is because of what is called "mutlitouch attribution", where it's important to know that the "foobar" campaign ought to receive credit even on the second visit. 

The reason why we are doing this custom is because we can link these associations at the analysis level, back in Big Query. We don't need this logic hardcoded. Eventually, the SDK itself will be making similar changes, but this is what we need right now for our own analysis.

junhouse (approved):
Thanks!

PR #742, opened 2023-12-01
Update ingestion.openapi.yaml

Tofufu:
For Ingestion API /engage Batch Profile updates: https://developer.mixpanel.com/reference/profile-batch-update

The example to show the $add operation has caused some confusions with users who are looking to set profile properties instead of incremental properties. 

Changing the example operations to $set instead.

Before:
Screenshot 2023-12-01 at 10 19 19 AM

After:
image


eerikson (approved):

mavlee:
this was resolved in a different PR

PR #749, opened 2023-12-05
Another hydration boundary fix

tiffanyqi:
- Testing the hypothesis that => breaks the hydration boundary
- For some reason, if you just do ### Mixpanel  and hard refresh, it breaks the page. no idea why.


pages/docs/features/custom-properties.mdx
```diff
@@ -124,7 +124,7 @@ Use the following functions in the **Formula** field to modify your custom prope
...
 | regex_match | Returns true if the pattern matches any part of the string. | regex_match(haystack, pattern)<br /><br />Example:<br />`regex_match("zzhaystackzz", "ha(..)ack")` -> true <br /> // Use (?-i) for case-sensitive matching: `regex_match("HAYSTACK", "(?-i)haystack")` -> false |
-| regex_replace | Replaces the parts of a string that match a regular expression with a different string. | regex_replace(haystack, pattern, replacement)<br /><br />Example:<br />  // convert currency string to number: `regex_replace("$1,234,567", "[^.0-9]\*", ""))` -> 1234567 |
+| regex_replace | Replaces the parts of a string that match a regular expression with a different string. | regex_replace(haystack, pattern, replacement)<br /><br />Example:<br />  // convert currency string to number: `regex_replace("$1,234,567", "[^.0-9]\*", ""))` -> `1234567` |
```
tiffanyqi, line 127: i think this thought that the number was a phone number in mobile safari

pages/docs/cohort-sync/integrations/segment.md
```diff
@@ -32,7 +32,7 @@ The connection requires a Segment Write Key. You can obtain this key via Segment
 
 ![Segment 3 Image](/segment3.png)
 
-### Mixpanel
+### Mixpanel Setup
```
tiffanyqi, line 35: this is crazy, for some reason if you just do ### Mixpanel and hard refresh the page, it breaks.

junhouse: I do see that if I go here https://docs.mixpanel.com/docs/cohort-sync/integrations/segment it just breaks. So I guess having more than one word for `###` fixes this issue for now?

junhouse: When I go to that link directly, I get the following error 
```
rollbar.min.js:1 TypeError: a.push is not a function
    at a.<computed> [as track_pageview] (<anonymous>:1:256)
    at p (_app-b863ce508cf6d63c.js:1:1095)
    at _app-b863ce508cf6d63c.js:1:4032
    at uI (framework-ce84985cd166733a.js:9:84130)
    at oU (framework-ce84985cd166733a.js:9:113201)
    at o (framework-ce84985cd166733a.js:9:107748)
    at x (framework-ce84985cd166733a.js:33:1373)
    at MessagePort.T (framework-ce84985cd166733a.js:33:1903)
```

junhouse: So I think when we get to that page via nav bar, since it's a single page app, we somehow don't trigger that `track_pageview`, but when we refresh, I think we are trying to call `track_pageview` but it's not ititialized?

junhouse (approved):

pages/docs/cohort-sync/integrations/segment.md
tiffanyqi: Yup, that's the error that this fix is attempting to fix. Just `# Mixpanel` breaks it and causes the error.

`track_pageview` should still be working for every page _except_ those with `# Mixpanel`... I agree that perhaps it's not initialized since the page is broken, but we still try calling `trackPageView` when we change pages.

PR #758, opened 2023-12-12
retention updates for upcoming changes

mavlee:


pages/docs/reports/retention.mdx
```diff
@@ -39,13 +38,13 @@ In this case, since we want to know retention after signup, choose "Sign Up" in
 
 ### Step 2: Choose Filters
 
-Filters exclude unwanted data. In this case, we only care about events performed on the iOS platform. Therefore, add a "Platform" filter, where Platform equals "iOS". At this point, your query should look like this:
+Filters exclude unwanted data. In this case, we only care about events performed on the iOS platform. Therefore, add a "Browser" filter, where Browser equals "Chrome". At this point, your query should look like this:
```
mataeuxg: This line doesn't really make sense here since we are switching from "platform" to "browser" yet still mention iOS platform.

PR #763, opened 2023-12-13
fix code example - removing "mixpaneltoken" from getGroup function

santigracia:
Based on:

https://mixpanel.github.io/mixpanel-react-native/Mixpanel.html#:~:text=getGroup(groupKey%2C%20groupID)

removing parameter MipanelToken


saminmadani (approved):
TIL!

PR #776, opened 2023-12-29
Create FAQs.md

vicentgilmixpanel:
Extended FAQs for billing & pricing created in collaboration with Finance, AEs, and Support.


pages/docs/pricing/FAQs.md
```diff
@@ -0,0 +1,146 @@
+# FAQs
```
myronkaifung, line 1: Could we update the page name to "Pricing FAQs" instead? This is to optimize the title for our indexing by our crawler.

PR #794, opened 2024-01-08
make `trackAutomaticEvents` as `false` by default

zihejia:
automatically collecting common mobile events is a legacy feature so make it `false` by default to be consistent.
reference: https://docs.mixpanel.com/docs/tracking-methods/sdks/android#legacy-automatically-tracked-events


saminmadani (approved):
lgtm

PR #806, opened 2024-01-11
Kurbycchua update property reference

kurbycchua:


kurbycchua:
Note to Reviewer:
1. Revamped the entire Property Reference section especially on the Default Properties to make it clearer and include other SDKs. Also added mention of support for Object and List of Objects data types and their limits
2. Added payload sample for List of Objects Property Support (Cart Analysis)

related slacks: 
1. On default props for other SDKs: https://mixpanel.slack.com/archives/CKGL4TYF3/p1704940486402089
2. On limits (testing) for Object and List of Objects: https://mixpanel.slack.com/archives/C04U70H63/p170532562414419

pages/docs/data-structure/property-reference.md
```diff
@@ -4,181 +4,127 @@
...
-| Reserved Properties                                                                                                                      | Mixpanel reserves certain property names (for [Events](https://help.mixpanel.com/hc/en-us/articles/360001355266#reserved-properties-for-events) and [User Profiles](https://help.mixpanel.com/hc/en-us/articles/115004708186#reserved-properties-for-user-profiles)) for special use cases, and these may or may not be automatically populated with values. The purpose of such Reserved Properties are for processing (ie event time stamping) or for specific system features (eg: cohort exports). Examples: `time`, `$email`, `$phone`, `$name`, `$created`. |
-
+| **Property Type** | **Description** |
+| ----------------- | --------------- |
+| [Event Properties](/docs/data-structure/events-and-properties) | Event Properties describes the events that are tracked within your product. |
+| [Super Properties](/docs/tracking-methods/sdks/javascript#super-properties) | Super Properties are a logical term for global Event Property that you can initialise to automatically attach to every subsequent Event you’re tracking at client side. |
```
argenisf: Hey Kurby, what do you think about wording it as: 

_properties that, after registration, are saved into the end user's device and appended to every event, from that device, after they have been registered and until persistence is cleared. This logical term is mostly applicable to our client-side libraries._

The mention on mostly applying to client-side libraries is just to be a bit more accurate since some server-side libraries, like at least the PHP one, have methods to register super props ([reference](https://github.com/mixpanel/mixpanel-php/blob/4b0fafacf2129eff5d50721e129b07f0c32687e7/lib/Mixpanel.php#L231)).

pages/docs/data-structure/property-reference.md
```diff
@@ -4,181 +4,127 @@
...
+| [User Profile Properties](/docs/data-structure/user-profiles) | User Profile Properties describe your users (they typically store current demographical information). User profiles are joined to events based on [Distinct ID](/docs/tracking-methods/identifying-users#what-is-distinct-id). |
```
argenisf, line 11: just making a note, not worth reviewing fully now, but when SCD is out, it's worth revisiting this definition

pages/docs/data-structure/property-reference.md
```diff
@@ -4,181 +4,127 @@
...
+| [Default Properties](/docs/data-structure/property-reference#default-properties) | Default Properties (for Events and User Profiles) are collected and populated with values automatically in Mixpanel. This can happen upon data ingestion or when using certain Mixpanel client-side SDKs. Typically, they have a dollar sign ($) or “mp\_” as prefix to distinguish them from normal properties. |
+| Reserved Properties | Mixpanel reserves certain property names (for [Events](/docs/data-structure/events-and-properties#reserved-event-properties) and [User Profiles](/docs/data-structure/user-profiles#reserved-user-properties)) for special use cases, and these may or may not be automatically populated with values. The purpose of such Reserved Properties are for processing (ie event time stamping) or for specific system features (eg: cohort exports). Examples: `time`, `$email`, `$phone`, `$name`, `$created`. |
  
 ## Default Properties
-Mixpanel's client SDKs automatically collect certain properties on every event. This document describes what those properties mean.
```
argenisf, line 19: I'd argue this paragraph could be removed as default properties are already defined in the table above.

pages/docs/data-structure/property-reference.md
```diff
@@ -4,181 +4,127 @@
...
-| $screen_width | Screen Width | Width, in pixels, of the device screen. |
-| $wifi | Wifi | Set to true if user’s device has an active, available Wifi connection, false if not. |
-| mp_processing_time_ms | Processing Time | UTC timestamp of when the event was processed by our servers. |
-| $mp_api_timestamp_ms | API Timestamp | Time at which the event was received by our API. |
-| $insert_id | Insert ID | A unique identifier for the event, used to deduplicate events that are accidentally sent multiple times. |
+| **Raw Name** | **Display Name** | **Description** | **[Javascript](/docs/tracking-methods/sdks/javascript)** | **[Android](/docs/tracking-methods/sdks/android)** | **iOS [Objective-C](/docs/tracking-methods/sdks/ios) / [Swift](/docs/tracking-methods/sdks/swift)** | **[Unity](/docs/tracking-methods/sdks/unity)** | 
+| -------- | ------------ | ----------- | :-----------: | :-----------: | :-----------: | :-----------: |
+| $device_id | Device ID | Autogenerated ID that is local to the device. Calling reset() resets this. More details [here](/docs/tracking-methods/identifying-users). | ✅ | ✅ | ✅ | ❌ |
```
argenisf: For the part "Calling reset() resets this", to avoid saying reset twice, perhaps we can say: Calling `reset()` regenerates this value.

pages/docs/data-structure/property-reference.md
```diff
@@ -4,181 +4,127 @@
...
+| $wifi | Wifi | Set to true if the user’s device has an active, available Wifi connection, false if not. | ❌ | ✅ | ✅ | ✅ | 
+| $bluetooth_enabled | Bluetooth | Set to true if Bluetooth is enabled, false if not. | ❌ | ✅ | ❌ | ❌ | 
+| $bluetooth_version | Bluetooth Version | Set to "none", "ble", or "classic". | ❌ | ✅ | ❌ | ❌ | 
+| $has_nfc | Has NFC | Set to true if Near Field Communication is being used, false if not. | ❌ | ✅ | ❌ | ❌ | 
+| $has_telephone | Has Telephone | Set to true if this device has telephone functionality, false if not. | ❌ | ✅ | ❌ | ❌ | 
+| $google_play_services | Google Play Services | Verifies that Google Play services is installed and enabled on this device, and that the version installed on this device is no older than the one required by this client. | ❌ | ✅ | ❌ | ❌ | 
 
 #### User Properties
 
 Do note that Mixpanel's default user properties are only updated when you send or update at least one property yourself.
```
argenisf: It'd be great to specify that it happens only when at least one property is updated via a $set or $set_once operation. Operations like $append or $union will not update these

pages/docs/data-structure/property-reference.md
```diff
@@ -208,3 +154,20 @@ Mixpanel supports five [data types](https://help.mixpanel.com/hc/en-us/articles/
 - A list of values as a JSON array e.g Favourite Genres = ["Folk","Alternative"] or Favourite Numbers = [1,5,10.0]
 - Limits of a List property: Event Property = 8KB, User Profile Property = 256KB, Each item in the list: 255 bytes
```
argenisf: We should correct this part from: "Each item in the list: 255 bytes"... to:

List of strings are also limited to 255 bytes per entry. List of objects have separate limitations covered below:

pages/docs/data-structure/property-reference.md
```diff
@@ -208,3 +154,20 @@ Mixpanel supports five [data types](https://help.mixpanel.com/hc/en-us/articles/
...
+
+### Object
+
+- Arbitrarily-nested groups of JSON key-value pairs e.g. Experiments = {"Exp Onboarding":"Quick","Exp Checkout":"Without Coupon"}
+- Limits of a Object property: Event Property = 8KB, User Profile Property = 256KB, max 255 keys per nested object with a max nesting depth of 3
+- Mainly supported in core reports (i.e. Insights, Funnels, Flows, Retention, Users / Cohorts, Events) as filters and breakdowns. Property Names (ie keys) within an object are not supported in Lexicon.
+
+### List of Objects
+
+- A JSON array of 1 level JSON objects with each object having similar set of key-value pairs e.g. Cart = [{"Brand":"Puma","Category":"Jacket","Price":30}, {"Brand":"Adidas","Category":"Hats","Price":15}]
```
argenisf: should be "similar set**s**" (plural)

argenisf (commented):
Hey Kurby, 

It looks great. I just had a few edits here in there.

pages/docs/data-structure/property-reference.md
kurbycchua: good suggestion, I made some minor edits to your suggested sentence, see if that works better

pages/docs/data-structure/property-reference.md
kurbycchua: yes agreed. with SCD I feel we should evolve how we document User Profile Properties - in that the SCD is totally like a separate entity / schema as I see it

pages/docs/data-structure/property-reference.md
kurbycchua: I'm retaining this as most often folks link directly to that sub-section on default properties thus it's good to have that info repeated plus the mention of CDP also mapping to our default props which I didn't want to mention in the above table so as to no over complicate the overview on different types of properties.

pages/docs/data-structure/property-reference.md
kurbycchua: thanks for this suggestion, updated

pages/docs/data-structure/property-reference.md
kurbycchua: updated, thanks for pointing that out. By the way does $add also count?

pages/docs/data-structure/property-reference.md
kurbycchua: good suggestion, reword and updated, see if it's good.

pages/docs/data-structure/property-reference.md
kurbycchua: haha good eyes! updated

pages/docs/data-structure/property-reference.md
argenisf: no, only $set and $ser_once. You can see here in the library how we do it for $set: https://github.com/mixpanel/mixpanel-js/blob/7d155e6fce7e59f3e13eb2e4d69b9905f10dbb36/dist/mixpanel.globals.js#L3204

but not for $add: https://github.com/mixpanel/mixpanel-js/blob/7d155e6fce7e59f3e13eb2e4d69b9905f10dbb36/dist/mixpanel.globals.js#L3287

argenisf (approved):

PR #815, opened 2024-01-17
Updating pages in relation to Migrating to Simplified ID Merge

kurbycchua:
1. Edited Migrating to Simplified ID Merge page for readability and better wording / positioning
2. Updated broken links due to Identifying Users being moved to a sub-section


pages/docs/tracking-methods/id-management/migrating-to-simplified-id-merge-system.md
```diff
@@ -133,180 +165,280 @@ Example 2:
...
+
+   - [Inviting users]([/docs/orgs-and-projects/roles-and-permissions#invite-users-to-a-project) to the project and granting them [Roles and Permissions](/docs/orgs-and-projects/roles-and-permissions#permissions)
+   - Creating [Teams](/docs/orgs-and-projects/roles-and-permissions#teams) and adding users to them
+   -  Adding [group keys](/docs/data-structure/advanced/group-analytics#group-keys-in-project-settings) from Group Analytics
+   -  Creating [data views](/docs/data-governance/data-views-and-classification#create-data-view) and adding users or teams to them
+   -  Creating or granting [service accounts](https://developer.mixpanel.com/reference/service-accounts) access to this project
+   -  Configuring [session settings](/docs/features/sessions)
+   
+   Note that the new project comes with newly generated project tokens, API secret, and potentially new sets of service account credentials. Replace the project tokens in your implementations with the new ones to start sending data to this new project. Ensure that any backend scripts using API secret or service accounts are updated as well.
```
wisey29: why do we add "potentially" here? The last I tested, only the org's service accounts are retained whereas the project's service accounts have to be recreated.

kurbycchua: yeah "potentially" mainly because of org level service accounts. If customers use org level accounts they would simply need to grant access to the new project instead of creating new ones. Also potentially refers to customers who also don't use service accounts and thus it should not matter to them. That's kinda why I used "potentially"

pages/docs/tracking-methods/id-management/migrating-to-simplified-id-merge-system.md
```diff
@@ -133,180 +165,280 @@ Example 2:
...
+
+If your historical events do not include both `$device_id` and `$user_id` that are required in Simplified ID Merge for identity merging, check if you can retrieve this ID mapping information from your system through other means. Instrument a dummy event that includes both `$device_id` and `$user_id` based on your ID mappings and send that to the new project to enable identity merging. You can choose any name for the dummy event except for \$identify, \$create_alias, and \$merge.
+
+##### Legacy ID Management
+
+If you had implemented using Mixpanel Client-Side SDKs (except for Unity) and have been making alias calls to link Anonymous ID to User ID, the Client-Side SDKs should have already populated `$device_id` and `$user_id` on your events (please verify this in your exiting Mixpanel project). These historical events can be directly imported into the new Simplified ID Merge project as they include reserved properties required for identity merging.
```
wisey29: typo "exiting" > "Existing"

pages/docs/tracking-methods/id-management/migrating-to-simplified-id-merge-system.md
```diff
@@ -133,180 +165,280 @@ Example 2:
...
 
-##### Migrating from Legacy ID Management
+1. <b>Cohorts, Custom Events, and Custom Properties</b>
+   
+   Manual Recreation: This involves manually copying and pasting / replicating the logic from the existing project into the new project's [cohorts](/docs/users/cohorts#creating-cohorts), [custom events](/docs/features/custom-events#create-a-custom-event), and [custom properties](/docs/features/custom-properties#creating-a-custom-property) definitions.
 
-If you are implementing via Mixpanel SDK and have been making an alias call to link anonymous ID to user ID, the SDK should have already populated `$device_id` and `$user_id` on your events (please verify this in your Mixpanel project). These historical events can be directly imported into Simplified ID project as they include reserved properties required for ID Merge to take place. 
+   For example for custom properties, follow steps to [Creating a Custom Property](/docs/features/custom-properties#creating-a-custom-property) and copy over the definition from the old project instead of starting from scratch.
```
wisey29: for some reasons, "copy over the definition from the old project instead of starting from scratch." gives me hope that I don't need to start from scratch - but the manual recreation is starting from scratch by itself, isn't it? maybe exclude "instead of starting from scratch."?

pages/docs/tracking-methods/id-management/migrating-to-simplified-id-merge-system.md
```diff
@@ -133,180 +165,280 @@ Example 2:
...
-3. In cases of a more intricate migration involving a larger data volume coming from different sources or posing a higher risk to the end-user experience, consider doing data backfilling before updating the live implementation. This approach allows for ample time to configure your new project, replicate reports and non-data entities in the new project against the backfilled data. While this may require multiple backfills, you have the option to deploy the live data implementation when you are ready. 
+   Before you move any Board, it's important to note the following:
+   - Duplicate the existing Board and move the new copy into the new project. This would minimise impact where users are still using Boards and reports in the old project.
+   - Any saved cohorts, custom events, custom properties, lookup tables would need to be created first as they don't get automatically moved as part of the Move Board.
```
wisey29, line 432: oh really? Is this confirmed? I can help test to validate coz I thought they will be brought over if they are used in reports.

pages/docs/tracking-methods/id-management/migrating-to-simplified-id-merge-system.md
```diff
@@ -133,180 +165,280 @@ Example 2:
...
-            "$user_id": "charlie"
-          }
-        }
-      ```
-   - Results of Simplified ID Merge:
+## Considerations when Migrating
+
+Take note of the following details when planning for the migration from Legacy ID Management or Original ID Merge to Simplified ID Merge:  
+
+1. Simplified ID Merge supports only one User ID (`$user_id`) per ID cluster as the canonical Distinct ID. If you need an ID management solution that supports multiple User IDs per user, such as both a email address and a phone number, it’s recommended to remain on Legacy ID Management or Original ID Merge which provide methods such as \$create_alias or \$merge to merge multiple User IDs.
```
wisey29: since you added canonical ID here, can we take this opportunity to indicate that user ID will become canonical ID? Users on Original ID will get that. Sth like "Simplified ID Merge supports only one User ID (`$user_id`) per ID cluster, and this User ID will serve as the user's canonical Distinct ID."

pages/docs/tracking-methods/id-management/migrating-to-simplified-id-merge-system.md
kurbycchua: good eyes.. updated

pages/docs/tracking-methods/id-management/migrating-to-simplified-id-merge-system.md
```diff
@@ -133,180 +165,280 @@ Example 2:
...
+     "event": "Add phone number as user ID",
+     "properties": {
+       "token": "{{token}}",
+       "$device_id": "+6512345678", 
+       "$user_id": "charlie"
+     }
+   }
+   ```
+      
+   - This results to phone number `+6512345678` being treated as one of the Anonymous IDs and prefixed with `$device:`. As such, if you subsequently send events with `$user_id` as `+6512345678` it will not be associated to `charlie` and would result to creating a completely different new ID cluster where `+6512345678` is the main User ID.
```
wisey29, line 199: @kurbycchua I can update the visual for this to include orphaned user - need?

pages/docs/tracking-methods/id-management/migrating-to-simplified-id-merge-system.md
kurbycchua: haha this was from your original which I tried to retain and did thought about removing it.. OK cool since you agreed with my previous intent I will remove it

pages/docs/tracking-methods/id-management/migrating-to-simplified-id-merge-system.md
kurbycchua: yes I had this issue before when I was creating board templates in the demo project. I had to result to NOT saving the custom props, cohorts, etc. It kinda makes sense since when you move board and their reports, the elements that a report refers to that are "saved" are mainly label pointers to the actual element since they are possibly being used across multiple reports, and can be updated independently, thus the report itself should not save anything beyond just the ID / label of those elements. So far only unsaved elements (customer props etc) are saved with the report itself as the actual logic - these get's ported over since they are part of the report.

kurbycchua: you can try testing is using a report that has a saved element vs a non-saved element

pages/docs/tracking-methods/id-management/migrating-to-simplified-id-merge-system.md
```diff
@@ -1,104 +1,134 @@
...
 
 ![image](/Tracking/project-setting.png)
 
-## Deciding When Migrating Makes Sense
-It is not possible to convert an existing project using Legacy/Original ID Merge to Simplified ID Merge. **To adopt Simplified ID Merge, you need to set up a new project from scratch**. This guide helps you in evaluating whether the migration will benefit your project based on your current ID management requirements and future plans. It outlines the pros and cons of each ID Management version and guides you through key considerations to make an informed decision. It also provides details on the resources required on your end should you decide to proceed with the migration. 
+## Deciding to Migrate
+
+It is currently not possible to automatically convert an existing project, already populated with data, on Legacy or Original ID Merge to Simplified ID Merge. This is because Simplified ID Merge has a very different backend architecture. <b>To adopt Simplified ID Merge, you would need to set up a new empty Mixpanel project</b>.
```
wisey29, line 28: hey kurby "This is because Simplified ID Merge has a very different backend architecture." - this is a very good addition because I've customers who just fixated on the fact that why CAN'T we do this for them, WHY isn't it built-in. They just don't understand. I'm wondering if you can kinda elaborate more on the WHY (I also don't mind just keeping it as it is).

pages/docs/tracking-methods/id-management/migrating-to-simplified-id-merge-system.md
```diff
@@ -1,104 +1,134 @@
...
+<br />
+
+The lack of a retroactive identity merging feature means that orphaned users are created whenever new Anonymous IDs are introduced during user interactions across multiple sessions, devices, and platforms. This prevents you from getting a holistic view of the user's journey. 
 
->When it’s ok to keep Legacy ID management: 
-If you are only tracking authenticated users (you do not track anonymous events), you don't need the retroactive ID Merge feature in Simplified ID Merge and should not consider the migration. We have preserved the documentation on the Legacy ID Management [here](https://github.com/mixpanel/docs/blob/main/legacy/aliases.md). 
+><b>Staying on Legacy ID Management</b> <br />
+> Note: If you are only tracking authenticated users (i.e. no tracked events while user is anonymous), you don't need the retroactive identity merging feature in Simplified ID Merge and should not consider the migration. We have preserved the documentation on the Legacy ID Management [here](https://github.com/mixpanel/docs/blob/main/legacy/aliases.md).
```
wisey29: (i.e. "do" no tracked events while user is anonymous)

pages/docs/tracking-methods/id-management/migrating-to-simplified-id-merge-system.md
kurbycchua: ok I like this change, updated

pages/docs/tracking-methods/id-management/migrating-to-simplified-id-merge-system.md
kurbycchua: @wisey29 yeah can you help update the diagram? Thanks!

pages/docs/tracking-methods/id-management/migrating-to-simplified-id-merge-system.md
kurbycchua: yeah I intentionally phrased that because of "I've customers who just fixated on the fact that why CAN'T we do this for them" but I didn't want to elaborate a bit more because it might invite more questions on "hey have you tried this or that to make it happen"

pages/docs/tracking-methods/id-management/migrating-to-simplified-id-merge-system.md
```diff
@@ -1,104 +1,134 @@
...
+
+The main limitation of Legacy ID Management was that anonymous user states could become orphaned. This happens when an anonymous user was initially tracked on one platform or device, signs up as a user, and later moved on to another platform or device, triggering various anonymous events before logging in. The anonymous events on the second platform would be orphaned, resulting in duplicated users on Mixpanel. 
+
+Aliasing on Legacy ID Management can only be done once, linking the identified User ID to the 1st Anonymous ID. Subsequent logins on a different platform or device where a user is identified again with the same User ID will tag subsequent events to the 1st Anonymous ID (from the 1st platform / device). Here’s a diagram illustrating how a typical user journey on different devices ends up creating an ophaned user.
```
wisey29: This sentence is a bit confusing for me " Subsequent logins on a different platform or device where a user is identified again with the same User ID will tag subsequent events to the 1st Anonymous ID (from the 1st platform / device). " 
Maybe we want to break them up to first emphasize 1) what happen to 2nd etc anonymous id 2) events with user id will still be linked to existing user? ("tag subsequent events to the 1st Anonymous ID" is confusing to me - maybe refer that as "an existing user" instead of "1st anon id"? )

pages/docs/tracking-methods/id-management/migrating-to-simplified-id-merge-system.md
kurbycchua: hmm not sure what you mean, I thought "no tracked events while user is anonymous" is an example to the statement "If you are only tracking authenticated users" maybe for clarity I updated it to "don't track events while user is anonymous"

pages/docs/tutorials/plan/setup.mdx
```diff
@@ -33,13 +33,13 @@ Click [here](https://mixpanel.com/settings/org/users-and-teams/find-your-org) to
...
 
 ![image](/tutorials/simplified-id-merge.png "Simplified ID Merge")
 
-We recommend using the **Simplied API** option as it is a straightforward simpler way of managing your users' identity in Mixpanel. More details about how Simplified API handles identity management can be found in our documentation under [Identifying Users](/docs/tracking-methods/identifying-users#how-does-the-simplified-api-differ-from-the-original-api). We will also cover a tutorial around this under the [Identity Management Best Practice](/docs/tutorials/implement/send-your-data#identity-management-best-practices) section.
+We recommend using the **Simplied API** option as it is a straightforward simpler way of managing your users' identity in Mixpanel. More details about how Simplified API handles identity management can be found in our documentation under [Identifying Users](/docs/tracking-methods/id-management/identifying-users#how-does-the-simplified-api-differ-from-the-original-api). We will also cover a tutorial around this under the [Identity Management Best Practice](/docs/tutorials/implement/send-your-data#identity-management-best-practices) section.
```
wisey29, line 42: straightforward "And" simpler way?

pages/docs/tracking-methods/id-management/migrating-to-simplified-id-merge-system.md
```diff
@@ -1,104 +1,134 @@
+# Migrating to Simplified ID Merge
+
 ## Overview
-Mixpanel has released three versions of ID management to date. Prior to March 2020, we have Legacy ID Management system.<br> 
-In March 2020, we released the Original ID Merge system which supports retroactive ID Merge. This allows the merging of events triggered both before and after authentication across multiple devices and platforms.<br>
-In March 2023, we released the current Simplified ID Merge system. It has a simpler implementation and doesn’t have the limitation of 500 IDs per ID cluster found in the Original ID Merge.<br> 
+Mixpanel currently has three versions of ID management:
+
+- Prior to March 2020, we only had 1 version that leverages primarily on aliasing to join the 1st anonymous state to the identified state. We will refer to this as **Legacy ID Management** (or Legacy for short).
+-  In March 2020, we released **Original ID Merge** (formerly known as just "ID Merge"), which supports retroactive identity merging by storing multiple Distinct IDs in an ID cluster. This enabled the merging of multiple anonymous states to an identified state across multiple devices and platforms.
```
wisey29: comments for legacy and original id merge - i think it's a bit technical to get into the concept of "aliasing" and "IDs" in our overview - can we explain legacy and original by spelling out their difference - mainly in terms of one can merge only first anon > identified conversion whereas the other can do many times (explaining them from the angle of user journey instead of underlying tech concept). Something like (feel free to edit): 

- Prior to March 2020, we only had 1 version that manage to merge the user's first time conversion from anonymous state to the identified state. We will refer to this as **Legacy ID Management** (or Legacy for short).
-  In March 2020, we released **Original ID Merge** (formerly known as just "ID Merge"), which supports retroactive identity merging, allowing the merging of anonymous states to an identified state across multiple sessions, devices or platforms.

pages/docs/tutorials/plan/setup.mdx
wisey29: oh sorry, I think I meant make this comment in my simplified id file.

pages/docs/tracking-methods/id-management/migrating-to-simplified-id-merge-system.md
```diff
@@ -1,104 +1,134 @@
...
+- **Original API**: Original ID Merge
+- **Simplified API**: Simplified ID Merge
 
-To determine your current ID Management version, navigate to Organization/Project Settings > Identity Merge. Organization Settings indicate the default version for every new project (Legacy ID Management is marked as "Disabled" there). **You can switch the ID Management version for a specific project via Project Settings provided no data has been ingested into the project.**
+![image](/Tracking/org-setting.png "Organization ID Merge Setting")
 
-![image](/Tracking/org-setting.png)
+<br />
+
+You can change the identity management version for a specific project (without affecting other projects) via Project Settings, provided no data has been ingested into the project. For new projects, we recommend setting the Simplied ID Merge (<b>Simplified API</b>) option as it is a straightforward simpler way of managing your users' identity in Mixpanel.
```
wisey29: straightforward "And" simpler way?

pages/docs/tracking-methods/id-management/migrating-to-simplified-id-merge-system.md
kurbycchua: Aliasing on Legacy ID Management can only be done once. Once a User ID is aliased to an Anonymous ID (typically on the 1st device where they started using your product), subsequent attempts to alias the same User ID to a different Anonymous ID (generate from a different platform or device) will fail. Here’s a diagram illustrating how a typical user journey on different devices ends up creating an ophaned user.

pages/docs/tutorials/plan/setup.mdx
kurbycchua: ok resolving it

pages/docs/tracking-methods/id-management/migrating-to-simplified-id-merge-system.md
kurbycchua: updated based on your suggestions and reworded a bit. Have a look and let me know

pages/docs/tracking-methods/id-management/migrating-to-simplified-id-merge-system.md
kurbycchua: I think I missed a word here and a comma, updated to "a generally more straightforward, simpler way"

pages/docs/tracking-methods/id-management/migrating-to-simplified-id-merge-system.md
wisey29: @kurbycchua got it - maybe sth like 
"... API secret, and new sets of service account credentials (if you create new service account in the project)" - so like just make sure that ppl give it a check.

pages/docs/tracking-methods/id-management/migrating-to-simplified-id-merge-system.md
wisey29: yes tested - the report looks the same on the surface but that's just label, the actual cohort, custom property etc is not carried over.

pages/docs/tracking-methods/id-management/migrating-to-simplified-id-merge-system.md
kurbycchua: ok reworded to "API secret, and new sets of service account credentials (if you had created new service account in the new project)."

I also caught "tokens" should be "token" since only 1 per project.

mherrman (approved):

PR #820, opened 2024-01-19
Update query.openapi.yaml

kurbycchua:
Update URL example for Query API since we change reports to be dependent on Boards that URL has already been outdated for a while


kurbycchua:
Since this is my first time update Dev Docs, please let me know if I did this change correctly

kurbycchua:
hey @mavlee @lindseybradford can one of you have a look to see if this update is correct and makes sense. Thanks!

lindseybradford:
> hey @mavlee @lindseybradford can one of you have a look to see if this update is correct and makes sense. Thanks!

I've also not done this before—deferring to @mavlee!

mavlee (approved):

PR #823, opened 2024-01-22
Update Events + User Properties Page

kurbycchua:
1. Revamp Event and Properties page to makes things clearer
2. Revamp User Properties page to makes things clearer
3. Updated Reserved Event + User Properties
4. Mention group profiles in User Profiles page
5. Added the following default props as well: $import, $watch_model, $duration


pages/docs/data-structure/events-and-properties.md
```diff
@@ -1,53 +1,100 @@
 # Events And Properties
 
+## Overview
 
+Events are the core of [Mixpanel's Data Model](/docs/tutorials/plan/tracking-strategy#the-mixpanel-data-model). All events should have an <b>Event Name</b>, a <b>Timestamp</b> of when that event has occured, and a <b>[Distinct ID](/docs/tracking-methods/id-management/identifying-users#what-is-distinct-id)</b> (Mixpanel's identifier for a user) to tie all events belonging to the same user. Events can optionally have a set of properties, which describes the event in more detail.
```
argenisf: On the distinct_id part, perhaps we can include a note or something on certain use cases in which a distinct_id is not needed, like importing ad cost... perhaps like: "to tie all events belonging to the same user (note there are a few use cases in which the event should not be tied to specific users, so they can lack a distinct_id, but this is the exception).

pages/docs/data-structure/events-and-properties.md
```diff
@@ -1,53 +1,100 @@
...
+## Reserved Event Properties
 
+Mixpanel reserves certain event property names for special processing or for specific system features. These properties, when populated, will affect the way Mixpanel processes your data.
+
+> <b>Note:</b> Also refer to the list of Mixpanel [Default Properties](/docs/data-structure/property-reference#default-properties) indicated in our documentation.
+
+| **Raw Name** | **Display Name** | **Description** |
+| ------------ | ---------------- | --------------- |
+| token | - | The project's token when sending data via [/track API](https://developer.mixpanel.com/reference/track-event). Not visible via Mixpanel UI reports. |
+| ip | - | The IP address that will resolve to `$city`, `$region`, and `mp_country_code` properties during ingestion of the event and thereafter discarded. By default, Mixpanel will pull ip from the source of the API request; if you don't want this resolved, set ip to "0". Refer to [Tracking Geolocation](/docs/best-practices/server-side-best-practices#tracking-geolocation-http-api) for example. |
```
argenisf: "By default, Mixpanel will pull ip from the source of the API request; if you don't want this resolved, set ip to "0".

Perhaps it should be worded as:

The `ip` event property is the value to be used for geo location parsing if `ip` as param to the /track endpoint has not been set to 1. Our client-side libraries, by default, will set the `ip` param to 1 so geo location is parsed from the incoming request, but [this can be disabled](/docs/privacy/protecting-user-data#disabling-geolocation) or, if you're implementing from the server, you can include the `ip` event property so it's parsed from it instead of the IP of the incoming request.

argenisf:
Hey @mavlee , I added a couple of comments as Kurby and I were chatting about this earlier, but I didn't get to do a comprehensive review, so it would still be good to review the rest of the changes.

pages/docs/data-structure/events-and-properties.md
kurbycchua: agree and updated with a few rewording

pages/docs/data-structure/events-and-properties.md
kurbycchua: agreed and updated with rewording

mavlee (approved):

PR #828, opened 2024-01-24
Restore masking effect on sidebar

eerikson:
Resolves [this task](https://mixpanel.atlassian.net/browse/INTRACTIVE-807).


junhouse (approved):

lindseybradford:
These vercel previews on blog and the mayven side are so helpful!

PR #860, opened 2024-02-07
Help doc links added to changelogs

peddemo-mixpanel:


mavlee (commented):
maybe we should standardize this message?

it would also be nice to embed the link in text and use relative paths

marissakuhrau (approved):

PR #870, opened 2024-02-16
Update JS SDK section on page views for single-page apps

chiyc:
## Description
The [latest release of the JS SDK](https://github.com/mixpanel/mixpanel-js/releases/tag/v2.49.0) adds the ability to automatically track page views in single-page applications. This updates the section on tracking page views to reflect the changes.

### DIFF
--- a/pages/docs/tracking-methods/sdks/javascript.md
+++ b/pages/docs/tracking-methods/sdks/javascript.md
@@ -17,7 +17,7 @@ mixpanel.track(
 );
 ```
 
-All events sent from the JavaScript library will be sent over HTTPS. 
+All events sent from the JavaScript library will be sent over HTTPS.
 
 ### Tracking Page Views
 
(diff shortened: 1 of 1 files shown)


pages/docs/tracking-methods/sdks/javascript.md
```diff
@@ -27,7 +27,20 @@ Page view tracking is turned off by default. Page view events can be added autom
 mixpanel.init('YOUR_TOKEN', {track_pageview: true});
 ```
 
-The `track_pageview` option currently does not auto-track page views in single-page applications. For single-page applications or other manual page view tracking, Mixpanel offers the standard page view event through `mixpanel.track_pageview()`. This call can be inserted into any hooks or listeners for your framework of choice.
+The default `track_pageview` setting does not auto-track page views in single-page applications. For manual page view tracking, Mixpanel can send the standard page view event through `mixpanel.track_pageview()`. This call can be inserted into any hooks or listeners for your framework of choice.
```
tdumitrescu: This paragraph doesn't make a lot of sense in its new form IMO. It used to be "we can't autotrack SPA, so here's how to do that manually or hook manual tracking to a SPA framework." Now it's "we don't autotrack SPA by default, and btw here's how to do manual tracking and how you would hook manual tracking to a SPA framework." It might be better to delete it altogether.

chiyc: That makes sense to me. Anyone that still needs to manually track probably knows how to do it. I removed the bit on manual tracking 👍

tdumitrescu (approved):

PR #895, opened 2024-03-01
LaunchDarkly integration docs

ranic:


pages/docs/tracking-methods/integrations/launchdarkly.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Exporting data from LaunchDarkly
+
+In order to perform its analyses, Mixpanel needs impression events. These events tell Mixpanel which user was exposed to which variant at which time. These events must look like this:
+
+```json
+{
+  "event": "$experiment_started",
+  "properties": {
+    "$user_id": "alice@example.com",
```
mavlee: shouldn't this be distinct id

pages/docs/tracking-methods/integrations/launchdarkly.md
```diff
@@ -0,0 +1,77 @@
...
+
+Doing this involves sending impression events from LaunchDarkly to Mixpanel and doing the analysis in Mixpanel.
+
+## Exporting data from LaunchDarkly
+
+In order to perform its analyses, Mixpanel needs impression events. These events tell Mixpanel which user was exposed to which variant at which time. These events must look like this:
+
+```json
+{
+  "event": "$experiment_started",
```
mavlee, line 18: i think you should be clear that this event and properties are special, this isn't a suggestion

pages/docs/tracking-methods/integrations/launchdarkly.md
ranic: With Simplified ID Merge we explicitly say "don't set distinct_id"

pages/docs/tracking-methods/integrations/launchdarkly.md
```diff
@@ -0,0 +1,79 @@
...
+
+```js
+function checkFeatureEnabled(experimentKey, defaultValue) {
+  // determine whether or not this feature is enabled for the current user
+  const variation = client.variation(experimentKey, defaultValue);
+
+  // This assumes the Mixpanel SDK is initialized
+  // Note: Use this precise event name
+  mixpanel.track("$experiment_started", {
+      Experiment name: experimentKey,
```
mavlee: these props need to be quoted

pages/docs/tracking-methods/integrations/launchdarkly.md
```diff
@@ -0,0 +1,79 @@
...
+    "time": 1709275888,
+    "Experiment name": "my_experiment",
+    "Variant name": "control"
+  }
+}
+```
+
+LaunchDarkly does not export this by default, so there are two options:
+1. Pay to [export](https://docs.launchdarkly.com/home/getting-started) from LaunchDarkly to Pubsub. If you go this route, you will need to write some code that pulls from PubSub and pushes to Mixpanel. We have a sample guide [here](/docs/tracking-methods/integrations/google-pubsub). This is server-side, so will likely be more accurate, but requires more work to integrate.
+2. Log assignments with wrapper code (see below). This is more convenient, but can be less reliable due to the limitations of client-side tracking.
```
mavlee: maybe make see below a heading link

pages/docs/tracking-methods/integrations/launchdarkly.md
```diff
@@ -0,0 +1,79 @@
...
+* Did users who were exposed convert better than users who didn't?
+* Were the right set of users exposed to my experiment?
+
+Doing this involves sending impression events from LaunchDarkly to Mixpanel and doing the analysis in Mixpanel.
+
+## Exporting data from LaunchDarkly
+
+In order to perform its analyses, Mixpanel needs impression events. These events tell Mixpanel which user was exposed to which variant at which time. 
+
+These events must be called `$experiment_started` and must have the `Experiment name` and `Variant name` properties:
```
mavlee: shouldn't end in : 
maybe also explain that this lets you automatically use experiment report

pages/docs/tracking-methods/integrations/launchdarkly.md
```diff
@@ -0,0 +1,79 @@
...
+```json
+{
+  "event": "$experiment_started",
+  "properties": {
+    "distinct_id": "alice@example.com",
```
mavlee, line 20: we don't have this or time inside properties right? this should be an actual event shape?

ranic: they are inside properties according to track/import api. which is unfortunate, but reality

PR #905, opened 2024-03-04
Update tealium.md

ak--47:
- fix casing + remove excessive formating
- add outline
- standardize language

### DIFF
--- a/pages/docs/tracking-methods/integrations/tealium.md
+++ b/pages/docs/tracking-methods/integrations/tealium.md
@@ -26,7 +26,7 @@ var utag_data = {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


pages/docs/tracking-methods/integrations/tealium.md
```diff
@@ -20,34 +26,33 @@ var utag_data = {
 }
 ```
 
-In order to *trigger* Mixpanel events based on the *values* of `tealium_event`s, you’ll need to create a **variable** in Tealium’s **data layer** for `tealium_event` (the *key* not the *value*).
+In order to trigger Mixpanel events based on the values of `tealium_event`s, you’ll need to create a variable in Tealium’sdata layer for `tealium_event` (the key not the value).
```
mavlee: missing space

mavlee: i'll just fix it

PR #906, opened 2024-03-04
Adding dynamic project token from url param projectToken

santigracia:
- found a way of not using Router and use the window.location searchParams

- needed to use useEffect because of the new tab functionality, switching between tabs changes the DOM and we need to update proj token accordingly

- based on current <code> element structure, I couldn't find an alternative to querySelectorAll().find to select that DOM element.

- Applied the code only to the needed SDKs sections: Javascript, Android, Swift, Obj-C according to onboarding guide here: https://mixpanel.com/report/project_id/view/view_id/setup/


tiffanyqi (approved):
i'd ideally like to see this in the components folder so that we can easily add and choose the specific string we want to replace so that we can move it elsewhere, but for now, this works for me. thank you!

saminmadani (approved):

myronkaifung:
Hi @santigracia , looks like the changes here were approved. Our team will be merging changes for smaller PRs moving forward. However, due to time elapsed, could you update the branch then merge the changes?

santigracia:
will create a new PR for this as this page has changed to:
https://docs.mixpanel.com/docs/quickstart/install-mixpanel

PR #907, opened 2024-03-04
Replace property value link looks with backslashes

tiffanyqi:
My guess on how to fix these hydration errors in mobile from: https://app.rollbar.com/a/mixpanel/fix/item/interactive-web/947#occurrences


mavlee:
this changes the look of the content
can we fix this without changing the content with escaping or something

mavlee (approved):

PR #910, opened 2024-03-06
docs: fix typo in Nth time example

cTxplorer:
The example screenshot shows the number of users that do Tutorial Complete 3 times (not 4 times)


ranic:
@cTxplorer thank you for making this fix!

PR #951, opened 2024-03-22
Update CODEOWNERS

ishamehramixpanel:


.github/CODEOWNERS
```diff
@@ -1,12 +1,3 @@
-# Default to engineering
+# Default to docs reviewers
 # ======================
-*                    @mixpanel/docs-engineering-reviewers @mixpanel/interactive
-package*.json        @mixpanel/interactive
```
eerikson, line 4: I'd still like to keep this one to engineering reviewers only if possible—it can be a little more dangerous in a way that's not limited to the scope of most PRs.

eerikson (approved):
I think if we can keep the package.json and package-lock.json files restricted to devs still, we should be fine on this. It's rarely updated anyway. Pre-approving.

austinpray-mixpanel (approved):

PR #974, opened 2024-04-04
Update javascript.md

sergiferr:
Adding the use case of having `persistence: 'localStorage'`which would block tracking across subdomains

### DIFF
--- a/pages/docs/tracking-methods/sdks/javascript.md
+++ b/pages/docs/tracking-methods/sdks/javascript.md
@@ -100,6 +100,8 @@ However, if your site is hosted on a domain like Heroku (or similar - [see a com
 
 For domains that don't allow cross-subdomain cookies, you should be setting `cross_subdomain_cookie: false`. Alternatively, you can also use a CNAME to change from yourdomain.herokuapp.com to yourdomain.com.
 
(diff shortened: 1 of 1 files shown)


pages/docs/tracking-methods/sdks/javascript.md
```diff
@@ -100,6 +100,8 @@ However, if your site is hosted on a domain like Heroku (or similar - [see a com
 
 For domains that don't allow cross-subdomain cookies, you should be setting `cross_subdomain_cookie: false`. Alternatively, you can also use a CNAME to change from yourdomain.herokuapp.com to yourdomain.com.
 
+Additionally, if the persistence setting is set to localStorage `persistence: 'localStorage'`, this will block tracking across subdomains. In that case, it should be modified to `persistence: 'cookie'`.
```
tdumitrescu, line 103: It sounds like
- using localStorage persistence _is_ a good solution for subdomain-hosted sites like the Heroku example above
- recommending cookie-persistence should _only_ be for the case where you need cross-subdomain tracking (that's virtually its only real advantage over localStorage in 2024)

myronkaifung:
Hi @sergiferr could you review Ted's comment?

PR #1012, opened 2024-04-18
Update JS docs for super property storage method

downeyfe:
The actual content for this section says that the super property info is stored in a cookie, so the heading is incorrect.

### DIFF
--- a/pages/docs/tracking-methods/sdks/javascript.md
+++ b/pages/docs/tracking-methods/sdks/javascript.md
@@ -147,7 +147,7 @@ If you want to store a super property only once (often for things like initial r
 
 This means that it's safe to call `mixpanel.register_once()` with the same property on every page load, and it will only set it if the super property doesn't exist.
 
-#### Super Properties Live in Local Storage
+#### Super Properties Live in a Cookie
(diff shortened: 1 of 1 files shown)


ranic:
@downeyfe thanks for the catch & the contribution!

pages/docs/tracking-methods/sdks/javascript.md
```diff
@@ -147,7 +147,7 @@ If you want to store a super property only once (often for things like initial r
 
 This means that it's safe to call `mixpanel.register_once()` with the same property on every page load, and it will only set it if the super property doesn't exist.
 
-#### Super Properties Live in Local Storage
+#### Super Properties Live in a Cookie
```
tdumitrescu, line 150: This is actually configurable in the SDK, so saying it's one or the other isn't really accurate. The default is cookie just for backwards compatibility, but we recommend initializing the SDK with localStorage super properties.

downeyfe: I see, good to know. In that case this entire paragraph should be updated to account for the different configurations

PR #1034, opened 2024-05-02
[Replay] JS SDK Documentation

jakewski:
Developer documentation for the Mixpanel Replay Beta


pages/docs/tracking-methods/sdks/javascript.md
```diff
@@ -368,5 +368,64 @@ mixpanel.init("<YOUR_PROJECT_TOKEN>", {api_host: "https://<YOUR_PROXY_DOMAIN>"})
...
+```javascript
+mixpanel.init(
+    "<YOUR_PROJECT_TOKEN>", 
+    {
+        record_sessions_percent: 1
+    }
+)
+```
+
+Start with a smaller percentage and tune to fit your analytics needs.
```
carlosdanieltt, line 388: Maybe the relationship between price and sampling percentage is obvious, but I would spell it out here.

pages/docs/tracking-methods/sdks/javascript.md
```diff
@@ -368,5 +368,64 @@ mixpanel.init("<YOUR_PROJECT_TOKEN>", {api_host: "https://<YOUR_PROXY_DOMAIN>"})
...
+
+```javascript
+mixpanel.stop_session_recording()
+```
+This will have no effect if there is no recording in progress.
+
+### Privacy
+
+#### User Data
+The Mixpanel SDK will always mask all inputs. By default, all text on a page will also be masked unless a different blacklist CSS selector is specified.
```
carlosdanieltt: It might be clearer to just use the record_mask_text_selector name to refer to it here?

carlosdanieltt (approved):
Looking good!  Just left some suggestions.

pages/docs/tracking-methods/sdks/javascript.md
jakewski: I think we're still pretty unclear on pricing since this is closed beta - I'll refrain from mentioning it for now

PR #1045, opened 2024-05-08
Add support for thumbnails and videos in changelog list. Also pagination.

mikemuzza:
This is a first attempt at bringing thumbnails to the list view of the changelog page. How to use it:

In the frontMatter section at the top of each new change log, either set a `thumbnail:` URL or a `video:` URL. The list view will pick this up and present it. I've set a few for the last 20 or so change log items. Here's an example:

`
video: "https://www.loom.com/embed/3407a37ce1634f51b94b380c40c96306"
`

I also used some chatgpt wizardry to add pagination, as loading too many loom embeds would slow things down. I'm not a web person and can't vet this, so it would be good to get some eyes on this portion especially.

Screenshot 2024-05-08 at 12 36 02 PM

### DIFF
--- a/components/changelog-index.js
+++ b/components/changelog-index.js
@@ -1,7 +1,39 @@
 import { getPagesUnderRoute } from "nextra/context";
+import ExtendedButton from '/components/ExtendedButton/ExtendedButton'
 import Link from "next/link";
 
-export default function ChangelogIndex({ more = "Read more" }) {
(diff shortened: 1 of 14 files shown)


ranic:
Removed a console.log and added a parameter to make the Loom embeds cleaner. Playing with the preview looks good. Merging it now, we can always fix issues forward later.

PR #1092, opened 2024-05-16
Move asset and add full length image for Home

spjtls9:


dongjae93 (approved):
LGTM. 1 super nit. The screenshot's side nav seems to have 2 of the same board. Can we edit it or something to just have one of them? It's possible they are 2 different boards but from gh preview they look the same

spjtls9:
> LGTM. 1 super nit. The screenshot's side nav seems to have 2 of the same board. Can we edit it or something to just have one of them? It's possible they are 2 different boards but from gh preview they look the same

hawkeye 👁️ working on it!

PR #1111, opened 2024-05-27
Update export.openapi.yaml

argenisf:
just making a change to the description to create a pending task to update our JS sample code for the raw export API. The current code does not work with the streaming result the API provides. Sample required code in Zendesk ticket 580207


mavlee:
for future reference, this should be done in issues

PR #1118, opened 2024-05-30
add CDP implementation snippets for SR

ak--47:
internal convo: https://mixpanel.slack.com/archives/C05PTEK8NQZ/p1717014118562549

RC with these methods: https://github.com/mixpanel/mixpanel-js-private/pull/202
  ^ this is already live on the CDN, so if the code looks ok, this should be good to go


ak--47:
thread with mParticle team: https://mixpanel.slack.com/archives/CKXQMCJ07/p1716389616967969

pages/docs/features/session-replay.md
```diff
@@ -83,6 +83,97 @@ Not currently, but we plan to add support soon.
...
+```javascript
+mixpanel.init('MIXPANEL-PROJECT-TOKEN', {
+	record_sessions_percent: 10,
+	loaded: function (mixpanel) {
+		//when mixpanel has loaded, check for mParticle device_id
+		const pollInterval = 500;
+		const maxAttempts = 10;
+		let attempts = 0;
+
+		const mParticleHasLoaded = setInterval(() => {
```
jakewski: can we use this `ready` function instead of polling? https://docs.mparticle.com/developers/sdk/web/initialization/#listen-for-sdk-initialization

ak--47: @jakewski strong agree that the `ready()` hook is better than polling;

if mParticle lib loads before mixpanel lib, than `ready()` would work fine.

but if mixpanel lib loads first, we can't pass anything to `window.mParticle.ready` because `mParticle` doesn't exist yet... so i don't see how we get around polling...

i did find the unminified version of their snippet:
https://github.com/mParticle/mparticle-web-sdk/blob/master/snippet.js

^ it looks like the stub out methods like we do an queue them up, but if a customer does:

```html
<head>
<script>
//mixpanel snippet
mixpanel.init()
</script>

<script>
//mparticle snippet
</script>
</head>
```

our ready code won't work; ideally it's one foolproof snippet that works in all cases... agree/disagree? (or am i missing something obvious)

ak--47: https://github.com/mixpanel/docs/pull/1118/commits/725bb4f87de1ba6610d9d98a687a4b50b5c5e306

jakewski (approved):
nice, thanks for putting this together!

PR #1120, opened 2024-05-30
Update event-approval.md

djmixpanel:
Adding update for new EA hidden events limits. (https://github.com/mixpanel/analytics/pull/63304)


themacguffinman (approved):

pages/docs/data-governance/event-approval.md
```diff
@@ -37,4 +37,6 @@ Events will be marked as "New" as long as the following criteria are met:
 If new property detection is enabled, notifications will also include a "View New Properties" button, which will take them to a Lexicon view of new properties.  Properties will be marked as "New" as long as the following criteria are met:
 - It was created within the last 30 days
```
kurbycchua, line 38: Find this weird though that it has only 1 bullet when we say "following criteria" maybe we should just say the following, unless we have other criteria?

If new property detection is enabled, notifications will also include a "View New Properties" button, which will take them to a Lexicon view of new properties. Properties will be marked as "New" as long as it was created within the last 30 days

pages/docs/data-governance/event-approval.md
```diff
@@ -37,4 +37,6 @@ Events will be marked as "New" as long as the following criteria are met:
...
 
-*Please note that all custom properties will **not** be marked as "New". 
+### Additional Notes
+- Please note that all custom properties will **not** be marked as "New"
```
kurbycchua: should we add a link to /docs/features/custom-properties for custom properties? and we should say like:

- Please note that all [custom properties](/docs/features/custom-properties) will **never** be marked as "New", since they are essentially formulas that transforms existing properties.

pages/docs/data-governance/event-approval.md
```diff
@@ -37,4 +37,6 @@ Events will be marked as "New" as long as the following criteria are met:
...
+- Event Approval is not available for projects with > 3000 hidden events, and will be auto-disable if projects exceed this limit
```
kurbycchua: minor nitpick, when we say "projects exceed" seems to come across to me as it might be a org level setting that project level setting

- Event Approval is not available for projects with > 3000 hidden events, and will be auto-disable if a project exceeds this limit

myronkaifung:
Hi @djmixpanel , would you please review the above, then merge the changes if everything is looking good?

PR #1129, opened 2024-06-05
Update retention.mdx

dorothy-koh:


Tofufu (commented):
Hi! ## looks like it might be too large (from looking at the preview), I believe this needs to be ####

myronkaifung:
Hi @dorothy-koh , would you please review the comment above?

PR #1140, opened 2024-06-11
Update legacy-mtu-billing.md

mherrman:


pages/docs/pricing/legacy-mtu-billing.md
```diff
@@ -43,7 +43,7 @@ Certain events and API updates are non-qualifying and excluded from the MTU calc
 
 While Mixpanel events generated by messages do not count toward your total MTU calculation, Message Open events do count towards MTU Guardrail.
 
-Similarly, while Mixpanel Warehouse driven Event Updates & Deletes and Profile Mirror Updates do not count towards your total MTU calculation, they do count towards the MTU Guardrail. 
+Similarly, while Mixpanel Warehouse driven Event Updates and Profile Mirror Updates do not count towards your total MTU calculation, they do count towards the MTU Guardrail.
```
argenisf, line 46: hey @mherrman, it looks like we are saying that DW deletes do not count towards the guardrail (it formerly said they did), but is that the case? my understanding was that both deletes and updates both are counted as one each towards the guardrail limit

PR #1150, opened 2024-06-13
use persistence cookie for subdomain tracking

junhouse:


tdumitrescu (commented):
you're going to have to do this in the webapp too, basically anyplace we call mixpanel.init for project 3

myronkaifung (approved):

PR #1151, opened 2024-06-13
Update Kapa module title and removing OAS files from search

myronkaifung:
Show "Mixpanel Ask AI" when using Ask AI tab, "Mixpanel Search" when using Search tab


junhouse (approved):
Thanks for making this change. I would wait for Isha to weigh in tab name changes

PR #1163, opened 2024-06-19
Add call out for ID management versions in identity API

myronkaifung:


reference/Identity API/identities/identity-create-alias.md
Tofufu, line 1: $create_alias can also be used for Legacy ID Management to alias identifiers together

myronkaifung: Thanks! Updated call out to reflect this

Tofufu (approved):
approved lgtm ty!

PR #1182, opened 2024-06-26
add security headers

junhouse:
Working as expected on devserver. Mostly same as what we use in `mixpanel.com`, but I added `kapa.ai` to the list to allow script loading. Also added `githubuser` to the list to allow img loading - this is used by Kapa.ai to load mixpanel logo on their widget.

Screen Shot 2024-06-26 at 12 57 35 PM


lindseybradford (approved):
lgtm

middleware.ts
```diff
@@ -0,0 +1,12 @@
...
...*.google.com.tw https://*.google.co.tz https://*.google.com.ua https://*.google.co.ug https://*.google.co.uk https://*.google.com.uy https://*.google.co.uz https://*.google.com.vc https://*.google.co.ve https://*.google.co.vi https://*.google.com.vn https://*.google.vu https://*.google.ws https://*.google.rs https://*.google.co.za https://*.google.co.zm https://*.google.co.zw https://*.google.cat; style-src 'self' 'unsafe-inline' https://cdn.mxpnl.com https://cdn-dev.mxpnl.com https://*.google.com https://*.marketo.com https://mixpanel.com https://*.mixpanel.com https://hello.myfonts.net; font-src 'self' data: https://cdn.mxpnl.com https://cdn-dev.mxpnl.com https://mixpanel.com https://*.mixpanel.com; frame-src 'self' https://js.stripe.com https://www.loom.com/embed/ https://player.vimeo.com/video/ https://www.youtube.com/embed/ https://*.facebook.com https://accounts.google.com https://static.addtoany.com https://*.bing.com https://*.doubleclick.net https://*.google.com https://*.marketo.com https://mixpanel.com https://*.mixpanel.com https://www.google.com/recaptcha/ https://recaptcha.google.com/recaptcha/ https://recaptcha.net https://*.recaptcha.net; worker-src 'self' blob:;`);
```
matthathornmp: Let's put this value in a constant at the least. I think we should also consider whether we want to import the CSP we use for the webapp, or whether it's better to have two separate policies. If the second, I think it would be good to limit ourselves to only those third parties that are needed for the blog alone. Do you have any thoughts on what we should prefer?

junhouse: Good idea but my reason for not using const was due to the fact it's only used here. Since this is a separate public repo - I don't think we can import from webapp. @scotmatson wanted to mostly copy over what we have in marketing-site here

PR #1183, opened 2024-06-26
Update billing faq

myronkaifung:


mherrman (commented):
I guess im wondering if we want to make this so transparent. Right now there is room ambiguity in how things are described which is what you are fixing but as like a business im wondering if documenting the behavior publicly is what we want

mherrman:
i guess im thinking maybe if an alternative might be to explicitly document this internally only... but would defer to product/opinions from others here

myronkaifung:
Let's remove this callout altogether so that we can control the narrative on this topic.

PR #1186, opened 2024-06-26
add new img src

junhouse:
adding t2.gstatic to support images within kapa widget


matthathornmp (approved):
LGTM. Unrelated, but do we need the 'unsafe-eval' and 'unsafe-inline' script sources?

junhouse:
> LGTM. Unrelated, but do we need the 'unsafe-eval' and 'unsafe-inline' script sources?

Let me double check!

PR #1205, opened 2024-07-02
Revert "continue on error (#1200)"

tiffanyqi:
This reverts commit 3dc49c80dc5a71d2edceee2f738cfff899274c83.

It's resolved (but still working with vercel to figure out how to fix), and would rather have at least one check here then both be off


matthathornmp (approved):
LGTM.

PR #1207, opened 2024-07-02
Add in archive for file uploads

tiffanyqi:
https://vercel.com/docs/cli/deploy#archive
Recommendation from Vercel to prevent uploading every file, upload it all together

After chatting with vercel support, they confirmed that outwardly there wouldn't be a difference in logs between an archived and non-archived build.


matthathornmp (approved):
LGTM.

PR #1211, opened 2024-07-03
Point User Profiles link to User Profiles page (instead of Events)

scott-marinoff:
"Learn more about profile properties" link pointed to the same page for 'event properties' link.

More helpful to send visitors to the User profiles page.


myronkaifung (approved):

kathmath (changes requested):
I think instead of _Learn more about profile properties_ and linking to reserved user props section makes sense to have _Learn more about user profiles_ and link to top level https://docs.mixpanel.com/docs/data-structure/user-profiles

@scott-marinoff what do you think?

kathmath:
looks like this was fixed with this update https://github.com/mixpanel/docs/pull/1224, going to close.

PR #1227, opened 2024-07-10
Set up spell checking in PRs

austinpray-mixpanel:
- Use the spell checker we use in the main monorepo. Works pretty good.
- Don't enable in PRs yet, just allow people to manually run it on main branch. So we can add obvious false positives to the known word list
- When fully deployed this will show inline comments about possible spelling suggestions. This is not configured to block PRs.

image


argenisf (approved):
Hey @austinpray-mixpanel ,

I'm afraid I'm out of my depth reviewing this but I trust you 😅 

Feel free to assign another reviewer on the eng side if you'd like

matthathornmp (approved):
LGTM, thanks Austin!

PR #1236, opened 2024-07-16
Update global filters section

aaqwu:
I believe there is a typo in the global filters section, but I'm not sure what the original sentence should have been. I added what I guess is correct based on the context.


Tofufu (approved):
Looks good! Thank you for editing :)!

myronkaifung (approved):

PR #1241, opened 2024-07-19
Updating images

dorothy-koh:
updating images under https://docs.mixpanel.com/docs/features/advanced#downloading-reports


argenisf:
hey @dorothy-koh it looks like the commit includes the new images, but they have not been replaced in the actual doc. I don't see the changes to the doc itself in the code, and when loading the preview, it also shows the old images.

mavlee:
this should be combined with #1242 

i'm going to close this

PR #1246, opened 2024-07-20
Update react-native.md

Tofufu:
Updating React Native SDK docs to include info about Expo Managed Workflow support (Javascript Mode).
Not sure if I should have this at the top or at the bottom? Thoughts?


mherrman:
i think at the top makes sense like you have it given this will be something customers would want to know early on.

mherrman (approved):

PR #1262, opened 2024-07-25
Updates on append mode and spelling corrections

argenisf:
Adding a callout on the possibility of missing records when including insert_time values in the past + a few spelling corrections


santigracia (commented):
LGTM! Approved!

santigracia (approved):

PR #1290, opened 2024-07-31
Myronkaifung identitymanagement update

myronkaifung:


myronkaifung:
tldr

* Overview page added to discuss what is distinct id, example of id management in action, and lightly touch on the 2 API versions
* Identifying users page now only discuss Simplified API, with a callout at the top to check versions and a link to the Original API doc. Focused on implementing and underlying mechanism and not just an overview
* New identifying users page focused only on Original API. Unindexed in the nav bar, only accessible via the call out from the main SImplified page. Focused on implementing and underlying mechanism and not just an overview
* updated redirect so any existing hyperlinks should continue to work

pages/docs/tracking-methods/id-management/identity-management.mdx
```diff
@@ -0,0 +1,107 @@
...
+  Already familiar with the concept of identity management? Click [here](/docs/tracking-methods/id-management/identifying-users-simplified) for a detailed implementation guide for Mixpanel ID management.
+</Callout>
+
+## Overview
+
+Identity management is the process of assigning IDs and maintaining ID mappings to ensure data are attributed to your users accurately. Mixpanel uses the **Distinct ID** property to identify your users and determine their uniqueness.
+
+## Distinct ID
+
+The Distinct ID is the identifier property that Mixpanel uses to keep identify your users. Every event data and profile update sent to Mixpanel should have a distinct_id property associated with it (except in the case of preventing [hot shards](/docs/debugging/distinct-id-limits)). The Distinct ID value tells Mixpanel which user to attach your events and profile property updates.
```
kurbycchua: For this sentence: "The Distinct ID is the identifier property that Mixpanel uses to keep identify your users." maybe we can reword to "Distinct ID is Mixpanel's identifier used to uniquely track a user in the system."

One note which I struggle as well when I update docs on distinct_id is that /import and /track uses distinct_id but /engage uses $distinct_id (as well as our UI). Which is why in the Reserved Event Properties page I decided to put both https://docs-git-myronkaifung-identitymanagement-update-mixpanel.vercel.app/docs/data-structure/events-and-properties#reserved-event-properties Not a big deal but just something to note. I think we can leave it as distinct_id

pages/docs/tracking-methods/id-management/identity-management.mdx
```diff
@@ -0,0 +1,107 @@
...
+</Callout>
+### Role of Distinct ID in Mixpanel
+
+Below are examples of Distinct ID's role in Mixpanel:
+
+**Insights report** - Distinct IDs are used to determine uniqueness, which is crucial for aggregation methods like Uniques and Average per User.
+
+**Funnels report** - Distinct IDs are used to follow a user through a conversion journey. If a user tracks events with inconsistent Distinct IDs between steps, they will be counted as a drop-off.
+
+**User Analytics** - Distinct IDs are used to index individual users. If a single users is tracked using 2 different Distinct IDs, they will appear as 2 different users in your project.
```
kurbycchua: In this case "User Analytics" is not really a report when compared to Insights and Funnels? I would suggest maybe changing this to say Users report to explain User Profiles where with 2 different Distinct IDs the same user would end up with 2 different profiles (which is quite common among many customers)? This paragraph is kinda a repeat anyway of what we already stated above in the call out on events having 2 different Distinct ID values.

pages/docs/tracking-methods/id-management/identity-management.mdx
```diff
@@ -0,0 +1,107 @@
...
+
+## Identity Management in Action
+
+Let's walkthrough how identity management affects your implementation using an example flow. Imagine you have a product that can be accesed from a website and a mobile app:
+
+### Without ID Management
+
+<img src="/no_id_management.png" alt="no_id_management" width="400"/>
```
kurbycchua, line 39: Would be nice if you can make the image background transparent so that it blends in with the doc background, looks more professional with diagrams. Let me know if you need help on this. Happy to do it.

pages/docs/tracking-methods/id-management/identity-management.mdx
```diff
@@ -0,0 +1,107 @@
...
+
+1. A user access your product on the website through a computer. The Javascript SDK automatically generates a Distinct ID for them, with the value of `123`.
+2. Later in the day, the same user access your product again, this time on their smart phone. The Android SDK automatically generates a Distinct ID for them, with the value of `ABC`.
+3. This single user would appear as two separate users in your Mixpanel project; once or their access through the website, and another for their access through your app.
```
kurbycchua: missing "f" in "once or their" or should be for

pages/docs/tracking-methods/id-management/identity-management.mdx
```diff
@@ -0,0 +1,107 @@
...
+
+Without identity management, a user who accesses your product from multiple device are incorrectly counted as two separate users.
+
+### With ID Management
+
+![with_id_management](/with_id_management.png)
```
kurbycchua, line 49: same comment for image background

pages/docs/tracking-methods/id-management/identity-management.mdx
```diff
@@ -0,0 +1,107 @@
...
+
+1. A user access your product on the website through a computer. The Javascript SDK automatically generates a Distinct ID for them, with the value of `123`.
+2. When the user registers for an account, your implementation make an identify call, using their name `charlie` as the User ID. This triggers an ID merge, resulting in an ID cluster containing both `charlier` and `123`, which will map to the same user in Mixpanel.
```
kurbycchua: this "your implementation make" I think should be "your implementation makes"

Also thinking instead of saying "using their name `charlie`" shall we say "using their login `charlie`" to avoid the notion of like using names but logins

this "containing both `charlier` and `123`" should be "containing both `charlie` and `123`" without the extra "r" at charlie

pages/docs/tracking-methods/id-management/identity-management.mdx
```diff
@@ -0,0 +1,107 @@
...
+3. Later in the day, the same user access your product again, this time on their smart phone. The Android SDK automatically generates a Distinct ID for them, with the value of `ABC`.
+4. When the user logs in to your product using the account they registered earlier, your implementation calls identify again using the same User ID. This triggers another merge, adding `ABC` to the existing ID cluster. Now `ABC`, `123`, and `charlie` are all in one cluster and maps to one user in Mixpanel.
+
+With identity management, a user who accesses your product from multiple devices are accurately resolved to a single user.
```
kurbycchua: I think "accesses" should just be "access"

pages/docs/tracking-methods/id-management/identity-management.mdx
```diff
@@ -0,0 +1,107 @@
...
+
+## Identity Merge APIs
+
+<Callout type="warning">
+  You cannot switch between the two APIs if your project already contains data in it.
+</Callout>
+
+Mixpanel supports 2 versions of the ID Merge API; Simplified ID Merge and Original ID Merge. Simplified ID Merge API is the default for new organizations.
```
kurbycchua: Suggestion:

Mixpanel supports 2 versions of the ID Merge API:
- Simplified ID Merge
- Original ID Merge

Simplified ID Merge API is the default for all new organizations created from April 2024.

pages/docs/tracking-methods/id-management/identity-management.mdx
```diff
@@ -0,0 +1,107 @@
...
+
+Due to the limited changes in functionality beyond a simpler implementation experience, we do not recommend that customers who have already implemented using the original API reimplement on the simplified API. Both APIs will remain supported and provide the same features.
```
kurbycchua: I know you copied this from the original docs, but thinking since this paragraph is quite confusion and contradictory where in we say "we do not recommend" yet we have docs on how to migrate to simplified, should we just say (from the last line above)

Simplified ID Merge API is the default for new organizations created from April 2024. Although it is not directly possible to switch from Original ID Merge (for a project that already has data) to Simplified ID Merge, both APIs will remain supported.

pages/docs/tracking-methods/id-management/identity-management.mdx
```diff
@@ -0,0 +1,107 @@
...
+
+You can set the default Identity Merge API version for your projects in the Organization Settings:
+
+![Untitled](/Tracking/id-merge-org-settings.png)
+
+For new projects without any data in it, you can switch the Identity Merge API version for the project in the Project Settings:
+
+![Untitled](/Tracking/id-merge-project-settings.png)
+
+Once there are any data in your project, your Identity API version cannot be changed.
```
kurbycchua, line 80: Instead of "Once there are any data in your project, your Identity API version cannot be changed." we can say "A project's Identity Merge API version setting cannot be changed once you start sending data into the project."

pages/docs/tracking-methods/id-management/identity-management.mdx
```diff
@@ -0,0 +1,107 @@
...
+**Simplified API**
+
+- Distinct IDs are determined using `$user_id`/`$device_id` in events.
+- The `$user_id` is always set as the distinct_id for the user.
+- There is no limit on the number of `$device_id`s that can be merged into a single `$user_id`.
+
+**Original API**
+
+- Distinct IDs are determined using the `$identify`, `$merge`, and `$create_alias` events.
+- The distinct_id is automatically determined by Mixpanel using one of the IDs in the cluster and is not user-configurable. Any IDs in the cluster can be used for ingestion.
```
kurbycchua, line 98: Just thinking if we should still mention this "Any IDs in the cluster can be used for ingestion." since it kinda implies that Simplified doesn't since we didn't mention it there.

pages/docs/tracking-methods/id-management/identity-management.mdx
```diff
@@ -0,0 +1,107 @@
...
+- There is a limit of 500 IDs that can be merged into a single ID cluster.
+
+### Third-Party Integration Support
+Customer Data Platform (CDP) integrations may require some configuration to work the Simplified API:
+
+- Segment works out of the box with both the simplified and original APIs with no special configurations.
+- Rudderstack has a [connection setting](https://www.rudderstack.com/docs/destinations/streaming-destinations/mixpanel/#connection-settings) that should match the API version configured on your Mixpanel project.
+- mParticle works out of the box with both the original API, and the Simplified API; when using the Simplified API (for new projects) ensure the "Simplified ID Merge" option is checked in mParticle's connection details
```
kurbycchua: capitalize "O" in "original API" since you capitalized "S" or maybe make it consistent with the rest of the bullets in this section where it's lowecase with the exception of "Simplified ID Merge" in quotes.

pages/docs/tracking-methods/id-management/identity-management.mdx
```diff
@@ -0,0 +1,107 @@
...
+Most other integrations are unaffected by this API change. These integrations are not involved in identity management, they send data to the ID they are given and will work the same way on the simplified API that they do on the original API.
```
kurbycchua: Not sure why in the preview the "Most other integrations..." paragraph got stuck to the previous bullet. Maybe because there is a missing period at the end of the last bullet?

This "will work the same way on the simplified API that they do on the original API." should be "will work the same way on the simplified API as they do on the original API."

pages/docs/tracking-methods/id-management/identifying-users-simplified.mdx
```diff
@@ -0,0 +1,255 @@
...
+
+This system is called ID Merge. In this guide, we walk through how to identify users in projects using the **Simplified ID Merge API** and exactly how it works under the hood.
+
+## Mechanism
+
+The **Simplified ID Merge API** uses the `$device_id` and `$user_id` event properties to determine the `distinct_id` that is set for your events.
+
+### distinct_id
+
+The `distinct_id` property (learn more [here](/docs/tracking-methods/id-management/overview#distinct-id)) is the identifier property that Mixpanel uses to determine the uniqueness of your users.
```
kurbycchua, line 33: should the link be docs/tracking-methods/id-management/identity-management#distinct-id

Also "The `distinct_id` property is the identifier property that Mixpanel uses to determine the uniqueness of your users." you can drop the 2nd "property" after identifier since you already said property just before

pages/docs/tracking-methods/id-management/identifying-users-simplified.mdx
```diff
@@ -0,0 +1,255 @@
...
+
+The `$device_id` event property is automatically generated by our client-side SDK and then attached to all of the user's events. All client-side events should contain a `$device_id`.
+
+If an event contains a `$device_id` without a `$user_id`, the value of the `$device_id` will be set as the `distinct_id` for that event.
+
+We can consider the `$device_id` as the anonymous ID for the user.
+
+### $user_id
+
+Calling `identify(<user_id>)` sets the `user_id` value passed to the identify function as the `$user_id` event property for the user's events moving forward.
```
kurbycchua, line 45: for this "Calling `identify(<user_id>)` sets" should we say "Calling `identify(<user_id>)` in Mixpanel SDKs sets" since in APIs identify does not exist for simplified

pages/docs/tracking-methods/id-management/identifying-users-simplified.mdx
```diff
@@ -0,0 +1,255 @@
...
+   
+4. This new user (U2) now logs in. Call `.identify(U2)`.
+       
+    | Event | $device_id | $user_id | distinct_id (set by Mixpanel) | Notes                      |
+    | ----- | ---------- | -------- | ----------------------------- | -------------------------- |        
+    | 1     | D1         |          | ~~*$device:D1*~~ ⇒ **U1**     |                            |
+    | 2     | D1         | U1       | U1                            | Links D1 ⇒ U1              |    
+    | 3     | D2         |          | ~~*$device:D2*~~ ⇒ **U2**     | Retroactively updated      |
+    | 4     | D2         |          | ~~*$device:D2*~~ ⇒ **U2**     | Retroactively updated      |
+    | 5     | D2         | U2       | U2                            | Links D2 ⇒ U2, D1, D2, and U1 are inside one ID cluster              |
```
kurbycchua, line 138: this is wrong "D1, D2, and U1 are inside one ID cluster" should be "D1 and U1 are inside one ID cluster and D2 and U2 are inside a totally separate cluster" Or just simply say "D2 and U2 are inside one ID cluster"

pages/docs/tracking-methods/id-management/identifying-users-simplified.mdx
```diff
@@ -0,0 +1,255 @@
...
+
+### Step 3 - Set the Authenticated ID once users log in
+
+Once the user logs in, you know their true ID, you should leverage the new ID for the user.
+
+Set the `$user_id` property to the Authenticated ID and continue setting `$device_id` to the Anonymous ID generated in step 1, including both the `$user_id` and `$device_id` in your events moving forward. 
+
+If Mixpanel receives an event with both `$device_id` and `$user_id` set, it will merge the two IDs together. This is essential to track pre-login and post-login behavior accurately. The `distinct_id` will be assumed to the `$user_id`.
+
+If you choose to manually define the `distinct_id` property, it should be the same value as the `$user_id`.
```
kurbycchua, line 164: hmm thinking maybe we should drop this line as it might create confusion?

"If you choose to manually define the `distinct_id` property, it should be the same value as the `$user_id`." since a reader might think they can set distinct_id = $device_id without the prefix which would end up a disaster

pages/docs/tracking-methods/id-management/identifying-users-simplified.mdx
```diff
@@ -0,0 +1,255 @@
...
+
+#### Example Python
+
+
+Here's a pseudocode example using Django's [cookies](https://django-book.readthedocs.io/en/latest/chapter14.html#cookies) and [authentication](https://django-book.readthedocs.io/en/latest/chapter14.html#using-users). It assumes the client is setting and sending cookies:
```
kurbycchua, line 169: I can't seem to access those links it's says 404 for both link

pages/docs/tracking-methods/id-management/identifying-users-simplified.mdx
```diff
@@ -0,0 +1,255 @@
...
+  # Note: leave the first argument blank here, since we're passing $device_id and $user_id as properties.
+  mp.track("", event_name, properties)
+
+def identify_user(request):
+  properties = {
+    "$device_id": uuid.uuid4(),
+    "$identified_id": request.user.username
+
+  }
+  track_to_mp(request, "$identify", properties)
```
kurbycchua, line 192: Hmm not quite sure about this code in simplified all identity events ($identify, $create_alias, $merge) will be ignored at ingestion as per my test previously thus sending $identify will not be processed. Did you try this code in a project with simplified ID merge? Maybe should change the event from "$identify" to just as "add_to_cart"

pages/docs/tracking-methods/id-management/identifying-users-simplified.mdx
```diff
@@ -0,0 +1,255 @@
...
+
+### Step 2 - Leverage Anonymous ID for anonymous events
+
+Set the `$device_id` event property to the anonymous ID you generated. You do not need to set the `distinct_id` property since it will be assumed to `$device_id` if there is no `$user_id` present in the event.
+
+### Step 3 - Set the Authenticated ID once users log in
+
+Once the user logs in, you know their true ID, you should leverage the new ID for the user.
+
+Set the `$user_id` property to the Authenticated ID and continue setting `$device_id` to the Anonymous ID generated in step 1, including both the `$user_id` and `$device_id` in your events moving forward.
```
kurbycchua, line 160: This is just a note, once $device_id and $user_id is merged, technically there is no need to keep sending $device_id in subsequent events and customer can just use $user_id. Thus you just need 1 event with both $device_id and $user_id to trigger the merge, any subsequent event while user is know can just be $user_id without the $device_id. Anyway just a note

pages/docs/tracking-methods/id-management/identifying-users-simplified.mdx
```diff
@@ -0,0 +1,255 @@
...
+  properties = {
+    "$device_id": uuid.uuid4(),
+    "$identified_id": request.user.username
+
+  }
+  track_to_mp(request, "$identify", properties)
+```
+## Best Practices
+
+**Call `.identify` upon a registration/login or when an app is re-opened in a logged-in state**
```
kurbycchua, line 196: It's important to note that for simplified, identify and reset only exists in client-side SDKs to set / reset $user_id in the cookie. Just thinking since we are in the midst if improving this, should we just state "Call `.identify` in Mixpanel's client-side SDK upon..."

pages/docs/tracking-methods/id-management/identifying-users-simplified.mdx
```diff
@@ -0,0 +1,255 @@
...
+
+By calling `.identify()` at these specific points in user journeys, you would be able to link the pre and post-login events to the same user on Mixpanel. Besides, calling `.identify` when the users re-open the app in a logged-in state ensures that all events in the session are tracked with the user's identifier such as user id.
+
+**Call `.reset` upon logout or when an app leaves an authenticated-state**
```
kurbycchua, line 200: Same comment as previous point on calling identify

pages/docs/tracking-methods/id-management/identifying-users-simplified.mdx
```diff
@@ -0,0 +1,255 @@
...
+
+By calling `.reset()` at logout or when a user enters an unauthenticated-state (such as timing out), you generate a new `$device_id` for your user, ensuring that multiple users that are sharing a single device would not be incorrect considered as a single user tracking events with the same IDs.
```
kurbycchua, line 202: the word "incorrect" should be "incorrectly"

pages/docs/tracking-methods/id-management/identifying-users-simplified.mdx
```diff
@@ -0,0 +1,255 @@
...
+
+**Track the unique identifier as a super property and user property to assist in troubleshooting**
+
+You can track the user's unique identifier as a super property via `.register()` and user property via `.people.set()` as soon as it is available in the app i.e. on a successful sign-up / login or when an app is re-opened in a logged in state. In the cases when ID Merge is not implemented properly, you can rely on these properties for troubleshooting purposes.
+
+See the SDK reference on [registering super properties](/docs/tracking-methods/sdks/javascript#super-properties) and [setting profile properties](docs/tracking-methods/sdks/javascript#setting-profile-properties).
```
kurbycchua: this "docs/tracking-methods/sdks/javascript#setting-profile-properties" should be "/docs/tracking-methods/sdks/javascript#setting-profile-properties"

The missing / infront will result to a broken link

pages/docs/tracking-methods/id-management/identifying-users-simplified.mdx
```diff
@@ -0,0 +1,255 @@
...
+
+**How does this relate to User Profiles?**
+
+[User Profiles](/docs/data-structure/user-profiles) are set directly on `$distinct_id`s, not on `$user_id`s or `$device_id`s. We recommend waiting until after a user is identified before setting user profile properties.
+
+It is possible to set user profile properties for unidentified users by sending the profile updates to `$distinct_id=$device:<device-id>`. However, user profile properties are not preserved when `$device_id`s are linked to `$user_id`s, so any properties set before the IDs became linked will need to be set again using `$distinct_id=<user-id>` once the user is identified.
+
+**Is it possible to merge two `$user_id`s?**
+
+It is not possible to merge 2 `$user_id`s together using the Simplified API. We don't recommend doing this in general, as it adds complexity to your identity resolution strategy. Instead, we recommend having a single, unchanging `$user_id` for each user and pointing all other IDs for that user to that single `$user_id`.
```
kurbycchua, line 246: I had plans to edit this line "Instead, we recommend having a single, unchanging `$user_id` for each user and pointing all other IDs for that user to that single `$user_id`" to

"Instead, we recommend having a single, unchanging `$user_id` for each user and pointing all other IDs (as $device_ids) for that user to that single `$user_id`."

See if it makes sense to you

pages/docs/tracking-methods/id-management/identifying-users-simplified.mdx
```diff
@@ -0,0 +1,255 @@
...
+
+**How should I link identified IDs from 3rd-party systems?**
+
+Attribution providers (like Appsflyer, Adjust, and Branch) use Mixpanel's SDK properly to set `$device_id` to whichever ID they use for attribution.
+
+For cohort syncs out to 3rd-party systems, we recommend designating a user property with the identifier of the user in that third-party system. More details are in our [integrations docs](/docs/cohort-sync/integrations); for example, see our [doc on exporting cohorts to Braze](/docs/cohort-sync/integrations/braze). If those integrations are bidirectional (eg: they send events back to Mixpanel), it's best to ensure that the user ID in both Mixpanel and the 3rd-party system is the same so that those events are sent to the correct user.
```
kurbycchua: The link /docs/cohort-sync/integrations/braze maybe can be more specifically link to /docs/cohort-sync/integrations/braze#matching-mixpanel-and-braze-users

pages/docs/tracking-methods/id-management/identifying-users-simplified.mdx
```diff
@@ -0,0 +1,255 @@
...
+
+**What is the status of Mixpanel's legacy `alias` method?**
+
+Prior to March 2020, the only way to connect users together was the `.alias()` method. This was very limited and was not retroactive; this meant that if a user used two devices and then logged in, you would lose activity for the user from one of the devices.
+
+If you set up Mixpanel prior to 2020, you may have implemented using the `.alias()` method. Alias is still supported in its original state and we have preserved its documentation [here](https://github.com/mixpanel/docs/blob/main/legacy/aliases.md), but if you want to revisit your identity management strategy, we recommend setting up a new Mixpanel project and using the best practices outlined in this guide.
+
```
kurbycchua, line 259: I noticed that the last FAQ in the original document that talks about $identity_failure_reason was not included. We should still include that as it is still valid for Simplified. You can copy paste below:

**Why is there a `$identity_failure_reason` and `$distinct_id_before_identity` in my event?**

The `$identity_failure_reason` property will be populated with a value `errAnonDistinctIdAssignedAlready` if the `$device_id` passed was already linked to another `$user_id`. The `$device_id` will be ignored and the `$user_id`, in the same event, will be used as the `distinct_id` for the event. 

The `$distinct_id_before_identity` property stores the original `distinct_id` (which was `$device:<value for $device_id>`) when the event was sent to Mixpanel before being mapped to the `$user_id`.

pages/docs/tracking-methods/id-management/identifying-users-original.mdx
```diff
@@ -0,0 +1,225 @@
+import { Callout } from 'nextra/components'
+
+# Identifying Users
```
kurbycchua: Just thinking if we should make this "Identifying Users (Original API)" reason is that as I was clicking on the callout links to go from Simplified to Original (and vice versa) the page was looking exactly the same to me at first look. I know the 2 pages are somewhat similar, but making the title heading more obvious might help before I start reading the content in case I was at the wrong page?

pages/docs/tracking-methods/id-management/identifying-users-simplified.mdx
```diff
@@ -0,0 +1,255 @@
+import { Callout } from 'nextra/components'
+
+# Identifying Users
```
kurbycchua: Just thinking if we should make this "Identifying Users (Simplified API)" reason is that as I was clicking on the callout links to go from Simplified to Original (and vice versa) the page was looking exactly the same to me at first look. I know the 2 pages are somewhat similar, but making the title heading more obvious might help before I start reading the content in case I was at the wrong page?

pages/docs/tracking-methods/id-management/identifying-users-original.mdx
```diff
@@ -0,0 +1,225 @@
...
+
+- What % of site visitors end up signing up?
+- How much of my Purchase revenue can I attribute to a particular campaign?
+- What is the conversion rate of reading a particular blog post -> signing up?
+
+This system is called ID Merge. In this guide, we walk through how to identify users in projects using the **Original ID Merge API** and exactly how it works under the hood.
+
+## Mechanism
+
+The **Original ID Merge API** uses three special events to link identities and set the `distinct_id` for your events; `$identify`, `$create_alias`, and `$merge` events.
```
kurbycchua: "The **Original ID Merge API** uses three special events to link identities and set the `distinct_id` for your events; `$identify`, `$create_alias`, and `$merge` events." 

some minor grammar corrections:

The **Original ID Merge API** uses three special events to link various user identifiers and sets the `distinct_id` for your events; these events are namely: `$identify`, `$create_alias`, and `$merge`.

pages/docs/tracking-methods/id-management/identifying-users-original.mdx
```diff
@@ -0,0 +1,225 @@
...
+"event": "$identify",
+      "properties": {
+          "$identified_id": "<user_id>",
+          "$anon_id": "<device_id>",
+          "token": "YOUR_PROJECT_TOKEN"
+      }
+}
+```
+
+The `$identify` event will only merge an `$anon_id` that is in UUIDv4 format (anonymous ID format generated by Mixpanel and Segment) and only if the $anon_id was not previously merged to another `$identified_id`.
```
kurbycchua, line 52: make $anon_id as `$anon_id` as I think without the ` it creates weird formatting with the $ sign

pages/docs/tracking-methods/id-management/identifying-users-original.mdx
```diff
@@ -0,0 +1,225 @@
...
+
+- something about identify event properties (failure reason, identified_id, id_before_identify, etc.)
```
kurbycchua: was this left so you could expand how to troubleshoot Original ID merge? Seems like a dangling sentence

pages/docs/tracking-methods/id-management/identifying-users-original.mdx
```diff
@@ -0,0 +1,225 @@
...
+
+### $create_alias
+
+The `$create_alias` event is used to create an alias. An alias is a new value that will be interpreted by Mixpanel as an existing value. This is triggered when calling `.alias` using the client-side SDKs.
```
kurbycchua: Hmm this sentence: "The `$create_alias` event is used to create an alias. An alias is a new value that will be interpreted by Mixpanel as an existing value. This is triggered when calling `.alias` using the client-side SDKs." does not make sense, maybe you can say 

The `$create_alias` event can be used to link two non-anonymous Distinct IDs. It creates an “alias” to another Distinct ID already present in the identity cluster and adds the same alias ID into the same cluster. This is triggered when calling `.alias` using the client-side SDKs.

refer to my this deck for inspiration: https://docs.google.com/presentation/d/1_4oKGzO3evq1yBKntyQUvBGcA-OcWDTy_oxQbgzxtV0/edit#slide=id.gf8d0d12897_0_3652

pages/docs/tracking-methods/id-management/identifying-users-original.mdx
```diff
@@ -0,0 +1,225 @@
...
+
+This is a legacy function for projects not using ID Merge. Learn more about the alias function [here](https://github.com/mixpanel/docs/blob/main/legacy/aliases.md).
+
+Here is a technical definition of the [$create_alias event in our API reference](https://developer.mixpanel.com/reference/identity-create-alias).
+
+### $merge
+
+The `$merge` event allows you to merge any 2 identifiers together. Unlike the `$identify` events, there are no payload requirements. As long as the merging of 2 IDs does not lead to an ID cluster that exceeds 500 IDs, it will be successful.
+
+This event is not triggered using any SDK functions; it can only be imported by calling the API directly. Learn more in our API reference [here](https://developer.mixpanel.com/reference/identity-merge).
```
kurbycchua: change: "This event is not triggered using any SDK functions; it can only be imported by calling the API directly. Learn more in our API reference [here](https://developer.mixpanel.com/reference/identity-merge)."

to: This event cannot be triggered using any SDK methods; it can only be processed when sent via the /import API directly. Learn more in our API reference [here](https://developer.mixpanel.com/reference/identity-merge).

pages/docs/tracking-methods/id-management/identifying-users-original.mdx
```diff
@@ -0,0 +1,225 @@
...
+      ]
+    }
+}
+```
+
+## Client-side Identity Management
+
+If using our Web/Mobile SDKs or a CDP like Segment or Rudderstack, there are only 2 steps:
+1. Call `.identify(<user_id>)` when a user signs up or logs in. Pass in the user's known identifier (eg: their ID from your database).
+2. Call `.reset()` when a user logs out.
```
kurbycchua, line 90: So in Original ID Merge, we generally do not recommend calling reset() since this would easily cause a user to hit 500 cluster, especially for websites / apps with timeout logouts. 

see: https://docs.google.com/presentation/d/1_4oKGzO3evq1yBKntyQUvBGcA-OcWDTy_oxQbgzxtV0/edit#slide=id.g106726c6eaa_0_60

pages/docs/tracking-methods/id-management/identifying-users-original.mdx
```diff
@@ -0,0 +1,225 @@
...
+
+- Any events prior to calling `.identify` are considered anonymous events. Mixpanel's SDKs will generate a `$device_id` and set that to `distinct_id` to associate these events to the same anonymous user. By calling `.identify(<user_id>)` when a user signs up or logs in, you're telling Mixpanel that `$device_id` belongs to a known user with ID `user_id`.
+
+- Under the hood, Mixpanel will stitch the event streams of those users together. This works even if a user has multiple anonymous sessions (eg: on desktop and mobile). As long as you always call `.identify` when the user logs in, all of that activity will be stitched together.
+
+- Calling `.reset` will clear the local storage (which contains the `$user_id` and `$device_id`), and generate a new `$device_id` for the session. It is recommended to call `.reset` when a user logout or times out of an authenticated session to prevent the unintentional merging of multiple users sharing one device.
```
kurbycchua: see comment above on reset(), best NOT to call reset, the only drawback of not calling reset is that any post-logout anonymous events will be linked to the previous user. If we want to recommend calling reset() I tend to recommend calling it when there is an explicit action to signup for a new account (ie clicking on Sign-up Button) so that any signup flow can be tied to the new user instead of the previous logged out user

pages/docs/tracking-methods/id-management/identifying-users-original.mdx
```diff
@@ -0,0 +1,225 @@
...
+
+<Callout type="warning">
+    The `distinct_id` is randomly selected by Mixpanel using the IDs inside of an identity cluster. This means that the canonical distinct_id could be set to a `$device_id` or your chosen `user_id`. This is not user-configurable. You can use any of the IDs in the cluster for ingestion, but only the canonical `distinct_id` is used in queries and exports.
```
kurbycchua: instead of saying "randomly" we should say "programatically" so something like:

The `distinct_id` is programatically selected by Mixpanel, using one of the IDs inside of an identity cluster, based on the most optimal merging process.

see: https://docs.google.com/presentation/d/1_4oKGzO3evq1yBKntyQUvBGcA-OcWDTy_oxQbgzxtV0/edit#slide=id.g1029aef1f24_0_399

Also this line "but only the canonical `distinct_id` is used in queries and exports." we can say "but only the canonical `distinct_id` can be used in queries and exports."

pages/docs/tracking-methods/id-management/identifying-users-original.mdx
```diff
@@ -0,0 +1,225 @@
...
+</Callout>
+
+## Example User Flows
+Let's walk through a few user flows where ID Merge is useful, and when to call `.identify()` and `.reset()` to use ID Merge properly.
```
kurbycchua, line 108: see comment on reset() maybe this is a good section to highlight when should reset be called to avoid hitting 500 limit

pages/docs/tracking-methods/id-management/identifying-users-original.mdx
```diff
@@ -0,0 +1,225 @@
...
+
+### New User Signup
```
kurbycchua, line 110: Would be nice to have the visual tables similar to the simplified ones. Happy to help with those.

pages/docs/tracking-methods/id-management/identifying-users-original.mdx
```diff
@@ -0,0 +1,225 @@
...
+
+1. A user lands in your product on a new device and interacts with your product before signing up. Our SDK will assign the user a random `$device_id` (D1) and persists it. All events tracked at this point will have `distinct_id` set to the `$device_id` value (D1).
+
+2. The user returns later and signs up for your product. You call `.identify(U1)`. Mixpanel adds U1 and D1 to the same identity cluster. Moving forward, any events tracked using either U1 or D1 will resolve to the same user. Mixpanel will set U1 or D1 as the `distinct_id` moving forward. This is randomly determined by Mixpanel and is not user-configurable.
```
kurbycchua: avoid the word "randomly" as per previous comment

pages/docs/tracking-methods/id-management/identifying-users-original.mdx
```diff
@@ -0,0 +1,225 @@
...
+    
+### Returning User
+
+1. The user from the previous flow returns, but is on a new device and has not logged in yet. Our SDK assign the user a random `$device_id` (D2) and persists it. All events tracked at this point will have `distinct_id` set to the `$device_id` value (D2).
```
kurbycchua: "Our SDK assigns" assign should be with "s"

pages/docs/tracking-methods/id-management/identifying-users-original.mdx
```diff
@@ -0,0 +1,225 @@
...
+    
+2. The user logs in and call `.identify(U1)` to tell us that the user on this device is the same user we have seen before. Mixpanel adds D2 to the identity cluster previous created (D1+U1). Moving forward, any events tracked using `distinct_id` U1, D1, or D2 will resolve to the same user.
```
kurbycchua: "The user logs in and call `.identify(U1)` to tell us.."

maybe as: "The user logs in and a call to `.identify(U1)` is made to tell us..."

pages/docs/tracking-methods/id-management/identifying-users-original.mdx
```diff
@@ -0,0 +1,225 @@
...
+
+### Multiple Users, One Device
+
+1. A first user begins using a new device. Our SDK will assign the user a random `$device_id` (D1) and persists it. All events tracked at this point will have `distinct_id` set to the `$device_id` value (D1).
+
+2. The user logs in and call `.identify(U1)`, which links the D1 and U1 to the same identity cluster.
```
kurbycchua: "The user logs in and call .identify(U1), which links.."

maybe as: "The user logs in and a call to .identify(U1) is made, which links..."

pages/docs/tracking-methods/id-management/identifying-users-original.mdx
```diff
@@ -0,0 +1,225 @@
...
+
+3. The user logs out. At this point, you should call `.reset()`, or manually generate a new `$device_id` if you are managing it yourself. A new user shows up and tracks events using this new `$device_id`, which is set as the `distinct_id` D2.
```
kurbycchua: We should perhaps completely reword expand this example to:

3. The user logs out. At this point, any events triggered will still be linked to U1. We don't call `.reset()` at this point since we do not know if this is the same user or a totally different user. Not calling `.reset()` at this point also prevents hitting the 500 ID cluster limit should the same user log back in there will be be new `$device_id` to link into the cluster.

4. A different new user shows up and clicks "Sign-up" button on the same device. Before triggering any event we call `.reset()` to force the clearing of `distinct_id` and to generate a new random `$device_id` (D2). This would help us to link any anonymous events as part of the sign-up flow to the new user instead of the previous logged out user.

5. This new user (U2) now completes sign-up and logs in as a separate user. Call `.identify(U2)` which will now link D2 and U2 to form one cluster. D1 and U1 remains as a separate cluster.

pages/docs/tracking-methods/id-management/identifying-users-original.mdx
```diff
@@ -0,0 +1,225 @@
...
+    
+4. This new user (U2) now logs in as a separate user. Call `.identify(U2)`. Now D2 and U2 are merged to form one cluster. D1 and U1 remains a separate cluster.
```
kurbycchua: see previous comment to change 4. to 5.

pages/docs/tracking-methods/id-management/identifying-users-original.mdx
```diff
@@ -0,0 +1,225 @@
...
+
+## Server-side Identity Management
+
+Our server libraries normally require that you specify the `distinct_id` value for each event. If you don't know the user's identity at the time the event is tracked, then they're an anonymous user. 
+
+When using our Web or Mobile SDKs, Mixpanel will automatically generate an `$device_id` that's local to that user's device, and set that value as the `distinct_id` for the event. This ID will persist on all events tracked by that user on that device, until you call `identify()` or `reset()`.
```
kurbycchua: "Mixpanel will automatically generate an `$device_id`" should be 

"Mixpanel will automatically generate a `$device_id`"

pages/docs/tracking-methods/id-management/identifying-users-simplified.mdx
```diff
@@ -0,0 +1,255 @@
...
+    | 2     | D1         | U1       | U1                            | Links D1 ⇒ U1              |    
+    | 3     | D2         |          | ~~*$device:D2*~~ ⇒ **U2**     | Retroactively updated      |
+    | 4     | D2         |          | ~~*$device:D2*~~ ⇒ **U2**     | Retroactively updated      |
+    | 5     | D2         | U2       | U2                            | Links D2 ⇒ U2, D1, D2, and U1 are inside one ID cluster              |
+
+## Server-side Identity Management
+
+Our server libraries normally require that you specify the `distinct_id` value for each event. If you don't know the user's identity at the time the event is tracked, then they're an anonymous user. 
+
+When using our Web or Mobile SDKs, Mixpanel will automatically generate an `$device_id` that's local to that user's device, and set that value as the `distinct_id` for the event. This ID will persist on all events tracked by that user on that device, until you call `identify()` or `reset()`.
```
kurbycchua, line 144: "Mixpanel will automatically generate an $device_id" should be

"Mixpanel will automatically generate a $device_id"

pages/docs/tracking-methods/id-management/identifying-users-original.mdx
```diff
@@ -0,0 +1,225 @@
...
+
+If you're tracking from servers, you'll need to generate and manage the IDs yourself and set it as the `distinct_id` directly
+
+### Step 1 - Generate an Anonymous ID
+
+The key is to have an ID that is unique to each user and persists during that user's session. We recommend generating a UUIDv4 and storing that value in a cookie. All common server frameworks provide a simple way to set and retrieve cookies per request.
+
+### Step 2 - Leverage Anonymous ID for anonymous events
+
+Set the `distinct_id` event property to the anonymous ID you generated. You do not need to set the `$device_id` property.
```
kurbycchua: suggestion: "You do not need to set the `$device_id` property."

maybe: "You do not need to set the `$device_id` property explicitly; although recommended should you want to track users using multiple or sharing devices"

pages/docs/tracking-methods/id-management/identifying-users-original.mdx
```diff
@@ -0,0 +1,225 @@
...
+
+### Step 3 - Set the Authenticated ID once users log in
+
+Once the user logs in, you know their true ID, you should leverage the new ID for the user.
+
+Merge the users together by importing the `$identify` event.
```
kurbycchua: sould we say "Merge the IDs together by sending the `$identify` event."

pages/docs/tracking-methods/id-management/identifying-users-original.mdx
```diff
@@ -0,0 +1,225 @@
...
+}
+```
+
+## Best Practices
+
+**Call `.identify` upon a registration/login or when an app is re-opened in a logged-in state**
+
+By calling `.identify()` at these specific points in user journeys, you would be able to link the pre and post-login events to the same user on Mixpanel. Besides, calling `.identify` when the users re-open the app in a logged-in state ensures that all events in the session are tracked with the user's identifier such as user id.
+
+**Call `.reset` upon logout or when an app leaves an authenticated-state**
```
kurbycchua: Maybe modify to: **Call `.reset` only when there is an explicit intent to change to a new user (ie clicking on Sign-up button)**

pages/docs/tracking-methods/id-management/identifying-users-original.mdx
```diff
@@ -0,0 +1,225 @@
...
+
+By calling `.reset()` at logout or when a user enters an unauthenticated-state (such as timing out), you generate a new `$device_id` for your user, ensuring that multiple users that are sharing a single device would not be incorrect considered as a single user tracking events with the same IDs.
```
kurbycchua: Maybe modify to: By calling `.reset()` only when there is an explicit intent to change to a new user, allows you to link any anonymous events, that may be important as part of a user journey, to the new user instead of the previously logged out user. This would also avoid unnecessarily hitting the 500 ID cluster limit when the same previous user simply logs back in since no new `$device_id` is generated.

pages/docs/tracking-methods/id-management/identifying-users-original.mdx
```diff
@@ -0,0 +1,225 @@
...
+
+**Track the unique identifier as a super property and user property to assist in troubleshooting**
+
+You can track the user's unique identifier as a super property via `.register()` and user property via `.people.set()` as soon as it is available in the app i.e. on a successful sign-up / login or when an app is re-opened in a logged in state. In the cases when ID Merge is not implemented properly, you can rely on these properties for troubleshooting purposes.
+
+See the SDK reference on [registering super properties](/docs/tracking-methods/sdks/javascript#super-properties) and [setting profile properties](docs/tracking-methods/sdks/javascript#setting-profile-properties).
```
kurbycchua: this "docs/tracking-methods/sdks/javascript#setting-profile-properties" should be "/docs/tracking-methods/sdks/javascript#setting-profile-properties"

The missing / infront will result to a broken link

pages/docs/tracking-methods/id-management/identifying-users-original.mdx
```diff
@@ -0,0 +1,225 @@
...
+
+**Keep a record of your ID management implementation**
+
+We encourage you to document your implementation (or create a diagram of the implementation). This will come in handy when you need to re-implement this on a new platform or troubleshoot ID management issue.
+
+## FAQ
+
+**Why is my user's distinct_id set to the anonymous device ID even after I identify them using my own user ID?**
+
+The canonical distinct_id is randomly selected by Mixpanel using any of the IDs inside of a cluster. This means that even after identifying your user, Mixpanel may choose your user ID or one of the other IDs inside of the cluster (such as the device ID) to serve as the displayed ID for the user. **This is random and not user-configurable.**
```
kurbycchua: 1. avoid using the word "random" as per previous comments.

pages/docs/tracking-methods/id-management/identifying-users-original.mdx
```diff
@@ -0,0 +1,225 @@
...
+
+Any IDs inside of the cluster can be used for ingestion and will resolve to the same user in Mixpanel, but only the canonical distinct_id is used in queries and exports.
```
kurbycchua: "but only the canonical distinct_id is used in queries and exports." to

"but only the canonical distinct_id can be used in queries and exports."

pages/docs/tracking-methods/id-management/identifying-users-original.mdx
```diff
@@ -0,0 +1,225 @@
...
+
+For debugging purposes, the Activity Feed view of a single user is updated in real-time (less than 1 minute delay). You can get to the Activity Feed by navigating to [Users](/docs/users/overview) and selecting a given user.
+
+It may take up to 24 hours for this mapping to propagate to all other parts of the system. This means that, in some cases, when analyzing a funnel that spans pre-login and post-login behavior in real-time, some may be shown as dropped-off, even though they've performed the conversion event.
+
+**How does this relate to User Profiles?**
+
+[User Profiles](/docs/data-structure/user-profiles) are set directly on `$distinct_id`s. We recommend waiting until after a user is identified before setting user profile properties.
+
+**Is it possible to merge two `$identified_id`s?**
```
kurbycchua: maybe change "`$identified_id`s" to just "Identified IDs" so as not to confuse with $identify's parameter

pages/docs/tracking-methods/id-management/identifying-users-original.mdx
```diff
@@ -0,0 +1,225 @@
...
+
+It is possible using the [$merge API event](https://developer.mixpanel.com/reference/identity-merge), which allows merging of any 2 ID clusters as long as it does not exceed 500 total IDs.
+
+**How should I link identified IDs from 3rd-party systems?**
+
+Attribution providers (like Appsflyer, Adjust, and Branch) use Mixpanel's SDK properly to set `$device_id` to whichever ID they use for attribution.
+
+For cohort syncs out to 3rd-party systems, we recommend designating a user property with the identifier of the user in that third-party system. More details are in our [integrations docs](/docs/cohort-sync/integrations); for example, see our [doc on exporting cohorts to Braze](/docs/cohort-sync/integrations/braze). If those integrations are bidirectional (eg: they send events back to Mixpanel), it's best to ensure that the user ID in both Mixpanel and the 3rd-party system is the same so that those events are sent to the correct user.
```
kurbycchua: The link /docs/cohort-sync/integrations/braze maybe can be more specifically link to /docs/cohort-sync/integrations/braze#matching-mixpanel-and-braze-users

kurbycchua (changes requested):
I have placed my suggestions and comments inline. Overall I like this new way of documenting ID Merge and also splitting out the Original vs Simplified as separate pages with Original being hidden. You actually beat me to it, since I had this in my backlog. Thus thank you for driving this and for asking me to review, now I can focus on other things.

I tried my best to put the same comments for the simplifies vs original pages which are somewhat similar, I might miss some things so do double check for comments on once side that might apply to the other. Once you're done happy to have a 2nd review before I approve it.

pages/docs/tracking-methods/id-management/identity-management.mdx
myronkaifung: Done. Making the callout about distinct_id vs $distinct_id instead.

pages/docs/tracking-methods/id-management/identifying-users-original.mdx
myronkaifung: I will add a callout for the reset and 500 limit, but going to leave the content the same since it is important to call reset at logout to prevent merging of the same multiple users on same device.

pages/docs/tracking-methods/id-management/identifying-users-original.mdx
myronkaifung: done using your recommended language in the slides

pages/docs/tracking-methods/id-management/identifying-users-original.mdx
myronkaifung: Done. Great suggestion

PR #1295, opened 2024-08-02
Update overview.mdx

Tofufu:
corrected pipeline addon call out + pipelines for older projs offset to utc and new projs should be exporting to utc as well


myronkaifung (approved):
thanks

PR #1304, opened 2024-08-05
feat: add warehouse connectors openapi

ytolstyk:


reference/Mixpanel APIs/overview.md
```diff
@@ -5,42 +5,55 @@ hidden: false
...
+
 **Standard Server:** `mixpanel.com/api/app/data-retrievals`
 **EU Residency Server:** `eu.mixpanel.com/api/app/data-retrievals`
 The retrieval and deletion API calls are updated for version 3 and are made for GDPR and CCPA compliance.
+
+### [**Warehouse Connectors API**](ref:warehouse-connectors-api)
+
+**Standard Server:** `mixpanel.com/api/app/projects`
+**EU Residency Server:** `eu.mixpanel.com/api/app/projects`
+Warehouse Connectors API
```
tiffanyqi: nit, might be better for a pm or something, but perhaps could have more info about what this does here rather than a repeat of the title?

reference/Warehouse Connectors API/overview.md
```diff
@@ -0,0 +1,14 @@
...
+createdAt: "2024-07-26T19:31:43.295Z"
+updatedAt: "2024-07-26T19:31:43.295Z"
+categorySlug: "warehouse-connectors-api"
+---
+
+Warehouse Connectors allow you to connect a warehouse of your choice to import events, users, groups, and lookup tables.
+
+The Mixpanel Warehouse Connectors API endpoints include:
+
+- Trigger a manual import sync
```
tiffanyqi: do we expect more public endpoints? if not, we don't need a bullet

reference/Warehouse Connectors API/overview.md
```diff
@@ -0,0 +1,14 @@
+---
+title: "Overview"
+slug: "warehouse-connectors-api"
+hidden: false
+createdAt: "2024-07-26T19:31:43.295Z"
+updatedAt: "2024-07-26T19:31:43.295Z"
```
tiffanyqi: can we update these created / updated values?

reference/Warehouse Connectors API/overview.md
tiffanyqi: or, we can remove lines 12 down because the navbar already has the info

tiffanyqi (approved):
code lgtm! only comments are around copy.
Also when you deploy to production, rearrange the nav bar so that "overview" is above "run an import" please, which you can do in the readme UI. thanks!

reference/Mixpanel APIs/overview.md
ytolstyk: fixed

reference/Warehouse Connectors API/overview.md
ytolstyk: fixed

reference/Warehouse Connectors API/overview.md
ytolstyk: updated

PR #1306, opened 2024-08-06
Update property-reference.md

kurbycchua:
Added reference to MaxMind as a reference when translating geolocation data from IP address. Updated a couple of pages to link back to this information

Slack Thread with Legal for reference: https://mixpanel.slack.com/archives/C6SNW4QKE/p1722995262139019


mavlee:
i'm not sure if we should / can mention this

will follow up with legal?

kurbycchua:
> i'm not sure if we should / can mention this
> 
> will follow up with legal?

@mavlee are you referring to the mentioning of MaxMind? I can ping legal if that helps just need to know which one you are referring to?

kathmath (approved):

PR #1317, opened 2024-08-07
Clarifying DW connector billing

argenisf:
Clarifying that updates/deletes are counted towards billing, even if the record updated is for a previous billing cycle.

I'm purposely not explicitly  indicating that new records for previous months are not billed. While true, it would be best to not advertise it directly


pages/docs/pricing/billing.md
```diff
@@ -189,7 +189,6 @@ We will only apply charges to event data points with timestamps within the curre
...
+New events imported through a warehouse connection are treated as any other source (same as importing through our SDKs or API directly). Warehouse data, similar to other sources, is billed in accordance with general Mixpanel pricing policies. 
 
+There is one additional field you will see in the Billing Page on your Data Usage Table called "Updates". Updates (updating an existing event or deleting an event) are generated by Mirror-type syncs and are tracked separately from event volume. Each update is counted as one event towards billing and they are counted and billed for the month they were triggered on, even if the record being updated is for a previous month. You can read more about these updates and how warehouse connectors may affect billing from the warehouse perspective [here](https://docs.mixpanel.com/docs/tracking-methods/data-warehouse#faq)
```
NealKaviratna: 🤔

Instead of clarifying here, might be worth making the change in the linked doc just because there is a bullet point there that captures the trickiness that updating an event you added new in the same month only counts as 1 event (and deleting one you just added counts as 0)

Just a thought!

argenisf: @NealKaviratna I think it's best if we call it out on both, so I added the clarification in the linked page too. With billing, it's always better to over communicate I think.

mavlee (approved):

pages/docs/pricing/billing.md
NealKaviratna: ooh yeah, good call!

PR #1340, opened 2024-08-12
Add geolocation page

myronkaifung:


pmackle (approved):

kathmath:
hey @myronkaifung @pmackle - i'm reviewing this PR from @kurbycchua that's related to geolocation. touching diff files but i think they should be reviewed together to make sure they're aligned. happy to review both but since y'all are already working on it can you coordinate? https://github.com/mixpanel/docs/pull/1306

myronkaifung:
Thanks for flagging @kathmath  Let me check in with Kurby. My perspective is that as long as there is no conflict, slight duplication of information is okay and better surface this info.

PR #1344, opened 2024-08-13
Documentation on removing SSO Configuration

pjboudrx:
We now allow organization administrators to remove SSO configuration, if necessary. This PR adds information about that operation to the SSO overview doc, and instructs customers who will need to use this in the Shared SSO doc.


myronkaifung (approved):
Just converted the shared-sso md to mdx so we can use callout components + minor edits.

PR #1363, opened 2024-08-20
Update SSO Overview

catliaw:
Added more information for JIT, cleaned up and clarified some text.


myronkaifung:
Hey @catliaw seems like the JIT photo was never uploaded. Would you please add it to /public/admin/sso ?

myronkaifung (approved):

PR #1366, opened 2024-08-20
Update overview.mdx

myronkaifung:


pmackle:
Is the plan to have all of the docs to have numbering for the FAQ sections?

For example [pricing](https://github.com/mixpanel/docs/blob/f318ec0cc5d904e915bdbaa5e7378b4b94c53e3b/pages/docs/pricing/billing.md?plain=1#L154), [webhooks](https://github.com/mixpanel/docs/blob/f318ec0cc5d904e915bdbaa5e7378b4b94c53e3b/pages/docs/cohort-sync/webhooks.md?plain=1#L130), and [boards](https://github.com/mixpanel/docs/blob/f318ec0cc5d904e915bdbaa5e7378b4b94c53e3b/pages/docs/boards/advanced.mdx?plain=1#L121) don't have numbering (though the way they have the headers/bolding differ). Might make more sense to adjust the header leveling instead for this doc so it is organized more similarly to the other Help Docs

myronkaifung:
@pmackle My goal overall with moving FAQ questions from headers is to remove clutter from the ToC on the right. For example in [this article](https://docs.mixpanel.com/docs/reports/funnels#faq), there is a separate header for each question which makes the ToC too long and doesn't provide a helpful summary at a glance.

Some what subjective depending on the number of headers already in the page

pmackle (approved):

PR #1374, opened 2024-08-21
Myronkaifung update segment content

myronkaifung:


myronkaifung:
Hi @Tofufu , would you please give this a passthrough if you have some time?

argenisf:
hey @myronkaifung it looks like because the file was renamed to .mdx, the changes look like deleting the old file and creating the new one with all new contents, so it's a bit hard to see what changed specifically. Are you looking for someone to proof read it or are you confident on the changes?

myronkaifung:
Hi @argenisf , I had Donna review the content for me before so I think it would be best for her to review since it isn't so easy to see the differences with the new filename here.

But a high-level summary:

focus more on Action
[Data Mapping section](https://docs-git-myronkaifung-update-segment-content-mixpanel.vercel.app/docs/tracking-methods/integrations/segment#data-mapping)
Updated [Group Analytics section](https://docs-git-myronkaifung-update-segment-content-mixpanel.vercel.app/docs/tracking-methods/integrations/segment#implementing-group-analytics-using-segment) to so people know how to set up mapping for group call

Tofufu:
https://docs-git-myronkaifung-update-segment-content-mixpanel.vercel.app/docs/tracking-methods/integrations/segment#3-track-user-actions

> Lastly, track your users' actions using Segment's track API call, which would trigger Mixpanel's track API and ingest an event in your project.

Event data is sent to Mixpanel's /import API endpoint for Segment Actions

Tofufu:
https://docs-git-myronkaifung-update-segment-content-mixpanel.vercel.app/docs/tracking-methods/integrations/segment#preset-actions

1. Since we are concentrating on Actions, for all the "Mixpanel Action" - Track Event links, we will want to have the links point to Ingestion API /import https://developer.mixpanel.com/reference/import-events
2. For the Mixpanel Action of "Identify Calls" - It's only linking to the Simplified ID Merge doc. We should include it's eqv to Identify User ($identify event if Original ID Merge) + Sets Profile Properties (Ingestion API /engage)

Tofufu:
https://docs-git-myronkaifung-update-segment-content-mixpanel.vercel.app/docs/tracking-methods/integrations/segment#implementing-group-analytics-using-segment

> Learn more about [Group Analytics here](https://docs-git-myronkaifung-update-segment-content-mixpanel.vercel.app/docs) and the Segment Group Spec here.

Group Analytics URL may be incorrect here

Tofufu:
https://docs-git-myronkaifung-update-segment-content-mixpanel.vercel.app/docs/tracking-methods/integrations/segment#faq

> 4. How do I track page views using Segment? Page calls are automatically tracked via Segment. These can be tracked from Segment to Mixpanel as Loaded a Page or Loaded a Screen. To turn them off, you need to configure this in your Segment settings.

With Page Calls, analytics.js collects it "by default" if users keep it as part of Segment's snippet (there is a `analytics.page()` after the `analytics.load()`): https://segment.com/docs/connections/spec/page/

Otherwise, customers have to call analytics.page(). We should probably rephrase this to page views can be tracked via analytics.page (per segment's doc it is typically part of the snippet). additionally, for actions - the default Event Name mapping for Page Calls is "Viewed _name_"
image
If users want a consolidated name e.g. "Page Viewed" or "Loaded a Page", they need to modify the Event Name mapping.

myronkaifung:
@Tofufu Just addressed the feedback you provided. Thank you!

Can you review for any additional feedback or approve the PR please?

pages/docs/tracking-methods/integrations/segment.mdx
Tofufu, line 1: Link is going to a "not found page". 

Looks like the Ingestion API /import endpoint links are missing an "s" at the end

Should be: https://developer.mixpanel.com/reference/import-events

Tofufu: outside of the above - everything else looks gtm ~ just need to fix the above links

Tofufu (approved):
lgtm!

PR #1396, opened 2024-09-02
GTM instructions for session replay

ak--47:
@jakewski @kurbycchua since we've be getting questions about GTM installation, i figured i'd add it here. i know tag managers are not CDPs, but this feels like the right place to put this.

### DIFF
--- a/pages/docs/session-replay/session-replay-web.md
+++ b/pages/docs/session-replay/session-replay-web.md
@@ -236,6 +236,26 @@ mixpanel.init('MIXPANEL-PROJECT-TOKEN', {
 });
 ```
 
+##### Google Tag Manager (GTM)
+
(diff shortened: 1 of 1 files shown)


argenisf:
@ak--47 @kurbycchua as a suggestion, why don't we create it as another FAQ? the CDP section is within the FAQ, under _Can I use Session Replay with a CDP?_ why don't we add another item as _Can I use Session Replay via Google Tag Manager (GTM)?_

ak--47:
@argenisf 

Candidly I was trying to avoid a whole separate section for tag managers, because tealium support is an unknown but if that more consistent with our style just do it.

I'm afk for a few days - are you able to make those edits so we can get it out the the people?

kurbycchua (approved):
I'm OK with this changes and I made the docs link relative instead of their full https links which is a best practice should we shift stuff easier to maintain

PR #1421, opened 2024-09-12
Design Changes

pmackle:
Current rough draft for the design changes. Still figuring out how to target the appropriate elements, this is just for showing work so far. Open to feedback while editing if anything sticks out as very wrong


pmackle:
Was actually a UI slip-up fixed by Tiffany here: https://github.com/mixpanel/docs/pull/1493

PR #1443, opened 2024-09-19
GOOT remove archive-tz to increase upload size

tiffanyqi:


austinpray-mixpanel (approved):
I guess this will fix the immediate failure but we will run into rate limit issues shortly

PR #1460, opened 2024-09-23
Update managing-projects.mdx

myronkaifung:
raw export date params should be in UTC regardless of project settings pre/post QTZ (SRFE-6468)

_Update: the above is out of date_


mherrman:
oh woah this is a pretty significant change in behavior for /export API no? old non-qtz projects used the project timezone and since we change to UTC this means you have to add time filters now. also these old projects stored the data in project tz so that complicates using those timestamps

mherrman:
or i guess if the only thing that changed is the validation on the date to be based on a UTC day then its probably ok...

myronkaifung:
Fact-checked. Date params for export API are either in UTC or current project timezone depending on creation date. Updated all references to timezones and exports to be consistent across docs and dev docs

mherrman (approved):

openapi/src/export.openapi.yaml
```diff
@@ -33,8 +33,20 @@ paths:
...
           description: >-
             Required if using service account to authenticate request.
-        - $ref: ./common/parameters.yaml#/query/fromDate
-        - $ref: ./common/parameters.yaml#/query/toDate
+        - in: query
+          name: from_date
+          schema:
+            type: string
+          description: >-
+            The date in yyyy-mm-dd format to begin querying from. This date is inclusive and interpreted as UTC timezone for projects created after 1 January 2023 and current project timezone for projects created before 11 January 2023.
```
howardchen11, line 41: @myronkaifung is this a typo "11 January 2023"

PR #1480, opened 2024-10-02
Update and rename php.md to php.mdx

myronkaifung:


myronkaifung:
There are 2 versions of the same page in this PR, `php.md` and `php.mdx`. The goal is to convert the file from MD to MDX, but Github treats this as a deletion/creation of a new file which makes it difficult to review.

For now, I am keeping both files in the branch so we can review the diff. Once the PR is approved, I will delete `php.md` before merging `php.mdx'.

PR #1482, opened 2024-10-03
Update session-replay-web.md - Segment snippet payload

santigracia:
payload.type is not returning the type of the payload, it needs to use payload.obj.type


Tofufu (approved):
lgtm ! looks corrected ~ when I initially did payload.type === 'track' (original docs) it wasn't adding the session props to my segment events and changing it to payload.obj.type dtrt
image

myronkaifung (approved):

PR #1485, opened 2024-10-03
[Docs] add sign up button to nav bar

ryancurtis1:
Adds a sign up button to the nav bar on docs, only when there is room

Also updates the background color of nav bar in dark mode per design

Example (Light):
Screenshot 2024-10-03 at 6 08 11 PM

Dark Mode:
Screenshot 2024-10-03 at 6 08 39 PM


components/SignUpButton/SignUpButton.tsx
```diff
@@ -0,0 +1,9 @@
+import style from "./SignUpButton.module.scss";
+
+export default function SignUpButton() {
+  return (
+    <div className={style.signUpButton}>
```
tiffanyqi: can we flip it so that `a` is outside of the div? i think that's more proper

tiffanyqi (commented):
do we also want a signup button in the sidenav if we get into mobile?

ryancurtis1:
On small screens:

< 400px:
Screenshot 2024-10-04 at 2 04 48 PM

400-767px:
Screenshot 2024-10-04 at 2 06 08 PM

768-800px:
Screenshot 2024-10-04 at 2 08 58 PM

myronkaifung (approved):

ryancurtis1:
> do we also want a signup button in the sidenav if we get into mobile?

Synced with Tara, and we will keep the sign up button in the nav bar for small

components/SignUpButton/SignUpButton.tsx
ryancurtis1: 👍

PR #1487, opened 2024-10-04
Convert to nextra 3

tiffanyqi:
Honestly, I was looking into [Nextra 3](https://the-guild.dev/blog/nextra-3) because of breadcrumb linking, however it turns out that a couple of things are improved if we do it this way.

# Pros
1. Increase in performance
2. Automatic breadcrumb linking (the first on the list)
3. Clearer code (example: sidebar icons moved directly to meta.ts rather than inside of theme.config)
4. Catches strange syntax things, for example weird image names

Before
Screenshot 2024-10-04 at 1 07 39 PM

After
Screenshot 2024-10-04 at 1 07 50 PM

# Changes necessary to use Nextra 3
(this is mostly covered per commit):
1. Package changes is the main reason the file is so big
2. Next.config.js -> next.config.mjs
3. Convert all _meta.json files to _meta.ts
4. Convert `Card` and `Tab` to `Cards.Card` and `Tabs.Tab`

I'm mostly annoyed that there are so many changes that are not retroactive, but at least it wasn't too bad


myronkaifung (approved):
We should prioritize fixing the code syntax highlighting since it makes the SDK pages a lot harder to read.

Not a blocker for this PR.

PR #1511, opened 2024-10-09
Update lookup-tables.md

dorothy-koh:
customer ticket: https://mixpanelsupport.zendesk.com/agent/tickets/613949 

slack confirming we do not support mapping lookup tables to the distinct_id prop: https://mixpanel.slack.com/archives/C03QFM9BL/p1728448901744569?thread_ts=1728381164.104809&cid=C03QFM9BL

*might need to update the phrasing later, _if_ Rajiv confirms that the behavior is for user profile properties only.


mavlee:
this should probably go in somewhere like FAQ

dorothy-koh:
sorry for the delay @mavlee updated it based on your feedback. Let me know what you think.

myronkaifung (approved):

PR #1525, opened 2024-10-10
[Docs] remove font weight hover effect for nav bar

ryancurtis1:
This was a request from design

### DIFF
--- a/pages/overrides.scss
+++ b/pages/overrides.scss
@@ -91,13 +91,6 @@ $border-radius-lg: 100px;
     font-size: 14px;
     line-height: 20px;
     padding: 8px;
-
-    //hover effect
-    & span:hover {
-      font-weight: 700;
-      // color: colors.$purple100;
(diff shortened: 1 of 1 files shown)


junhouse:
Can you add before and after? Navbar seems like it's the same as prod?

ryancurtis1:
> Can you add before and after? Navbar seems like it's the same as prod?

Oh you're right. It looks like the hover effect is already removed from prod. I think the upgrade to nextra 3 changed the links from having nested `span` elements inside the `<a>` tags for each link to just an `<a>` so now the code isn't doing anything in prod since the css selector is looking for a span. Would still be good to clean up the css though.

It makes me wonder if there's any other parts of the css overrides that are no longer working as intended 🤔 cc @tiffany-mixpanel 

Before nextra3:
Screenshot 2024-10-10 at 1 12 54 PM

junhouse (approved):
Thanks!

PR #1542, opened 2024-10-15
update gdpr openapi spec

myronkaifung:


myronkaifung:
Tested all of the endpoint and it works. Response payload matches example payloads

PR #1560, opened 2024-10-22
Update funnels.mdx

dorothy-koh:
customer was confused about how exclusion step works 
reference ticket: https://mixpanelsupport.zendesk.com/agent/tickets/615806


pages/docs/reports/funnels.mdx
```diff
@@ -569,3 +569,14 @@ To calculate the actual p-value, we estimate the hypergeometric cumulative distr
 In the CDF, the value of any point (X) represents the probability that a random draw would result in fewer conversions P(k < X). 1 - P(k < X) represents the probability that a random draw would result in more conversions P(k >= X).
 
 These two probabilities are used to represent the probability that the selected segment will either outperform (P(k < X)) or underperform (P(k > X)) the overall set of users. Mixpanel takes the higher probability of the two, and calculate the p-value as 1 - max(P(outperform), P(underperform)).
+
+### Can I use the Exclusion Step feature to count each entry into the Funnel for Total Conversions?
```
myronkaifung: Hi @dorothy-koh , looks like this information is already documented under Exclusion Step. (specifically [here](https://docs.mixpanel.com/docs/reports/funnels#exclusion-steps-exclude-users-who-did:~:text=An%20exclusion%20step%20can%20be%20placed%20between%20any%20steps%20in%20the%20funnel.%20It%20cannot%20be%20the%20first%20or%20last%20step.))

Perhaps we can reword the existing note about Exclusion step not being used for first step and last step to make it more clear instead of adding it as a FAQ which adds length to the already lengthy page.

dorothy-koh:
oh I thought the [note](https://docs.mixpanel.com/docs/reports/funnels#exclusion-steps-exclude-users-who-did) you highlighted is regarding _how_ to build a funnel, rather than the _expected_ results of building the funnel. 

I could include it under the same section (since they both concern exclusion steps), but I'm concerned that having multiple examples might confuse the reader.

myronkaifung (approved):

PR #1574, opened 2024-10-25
[Changelog] post page updates to use new formatting

ryancurtis1:
a follow up PR will add the new formatting to more than the one post

Example Post Page:
Screenshot 2024-10-25 at 4 59 47 PM
Screenshot 2024-10-25 at 4 59 54 PM


pages/changelogs/_meta.js
```diff
@@ -0,0 +1,8 @@
+export default {
```
junhouse, line 1: what does this do?

junhouse (approved):
lgtm

pages/changelogs/_meta.js
ryancurtis1: This just allows us to use the default nextra layout on the individual post pages as opposed to the full screen layout I added for the main changelog page

PR #1575, opened 2024-10-25
[Changelog] update formatting for past 10 posts to match new styling

ryancurtis1:


pages/changelogs/2024-05-09-home.mdx
```diff
@@ -6,9 +6,14 @@ createdAt: "2024-05-09T16:50:45.397Z"
...
 ---
 
-![Home](/changelog/home.png)
+<ChangelogPostHeader
+  date="2024-05-09"
+  image="/changelog/home.png"
+  title="Home"
+/>
 
 Home is the new landing experience in Mixpanel. It’s for you to jump back into your workflows and get a glimpse into what’s happening across your product and your project in Mixpanel.
```
tiffanyqi, line 19: do you need both of these? can we make it so that we only need one? (as in description and date for example only need to be written once rather than copy pasted, hard to maintain)

ryancurtis1: So on the /changelog page in the feed of posts, the content relies entirely on the frontmatter (the context between the `---`). On the individual pages I couldn't figure out how to add a header that could use the frontmatter in a not super hacky way which is why I added this new component. 

I agree it's not ideal that the content is required in two places but hopefully maintenance is not too bad if PMs are just copy pasting previous pages when creating their new updates. But any ideas or insights to how nextra works?

junhouse: is it possible to use a `variable` here?

ryancurtis1: It seems like using a variable here isn't supported by markdown 🤔 

Let me check with Isha to see what she thinks about this format since it will be mostly PMs adding new posts

ryancurtis1: Checked with Isha and this should be fine if we don't see any better solutions currently since the PMs will just be copy and pasting when creating a new post

tiffanyqi (approved):

PR #1582, opened 2024-10-29
[Changelog] add subscribe to product updates modal and functionality

ryancurtis1:
Adds a modal with subscription form similar to what is on the What's New Page and sends to the same gtm_form but gets a 400 error because the preview links for docs are not on a mixpanel domain

Screenshot 2024-10-30 at 2 49 16 PM

### DIFF
--- a/components/SubscribeButtonWithModal/SubscribeButtonWithModal.tsx
+++ b/components/SubscribeButtonWithModal/SubscribeButtonWithModal.tsx
@@ -1,8 +1,10 @@
 import { useState, MouseEvent } from "react";
-import { Checkbox, Dialog } from "@headlessui/react";
+import { Dialog } from "@headlessui/react";
+// https://www.tailwind-variants.org/docs
+import { tv } from "tailwind-variants";
+
 import { track } from "../../utils/tracking";
 import { APIMethods, BaseAPI, HeaderContentType } from "../../utils/client";
-//import { APIMethods, BaseAPI, HeaderContentType } from "@src/utils/client";
 
 // default, focus, error, success
 const enum SubscribeInputState {
@@ -27,6 +29,8 @@ export default function SubscribeButtonWithModal() {
   const [subscribeInputState, setSubscribeInputState] = useState(
     SubscribeInputState.Default
   );
+  const submitDisabled =
+    subscribeInputState === SubscribeInputState.Error || !agreedToTerms;
 
   const validateEmail = (): string => {
     let errorMsg = ``;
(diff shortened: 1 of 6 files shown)


components/SubscribeButtonWithModal/SubscribeButtonWithModal.tsx
```diff
@@ -0,0 +1,198 @@
...
+const enum SubscribeInputState {
+  Default = `default`,
+  Focus = `focus`,
+  Error = `error`,
+  Success = `success`,
+}
+
+// From /iron/common/util/string.ts
+// See https://stackoverflow.com/questions/46155/how-can-you-validate-an-email-address-in-javascript
+export const IRON_EMAIL_REGEX =
```
junhouse, line 20: Double check w/ marketing to see if they want to use list of domains that they don't want. e.g. gmail.com. If so, we need to copy over the list of domains to `docs` page as well. But it may be weird to push list of domains we don't want to a public repo - might be better to do it in backend. Just send 200 and don't ingest the email or sth.. But yea, I would first check w/ Allie.

components/SubscribeButtonWithModal/SubscribeButtonWithModal.tsx
```diff
@@ -0,0 +1,198 @@
...
+                ? // @ts-ignore
+                  mixpanel?.get_property(`distinct_id`)
+                : ``,
+          },
+        });
+        track(`[DOCS] Subscribed to Product Updates`, {});
+        setSubscribeInputState(SubscribeInputState.Success);
+        setError(``);
+      } catch (e) {
+        // send to rollbar.
```
junhouse, line 80: Do we have a rollbar instance set up in `docs`? I think we do.. Could you also remove the commented out blog tracking code here.

junhouse (approved):
Left a few requests.

components/SubscribeButtonWithModal/SubscribeButtonWithModal.tsx
ryancurtis1: Copied over the tracking utils we have to send errors to rollbar. Unless there's any reason we wouldn't want those in docs?

components/SubscribeButtonWithModal/SubscribeButtonWithModal.tsx
ryancurtis1: Checked with Allie and we don't want to limit to company emails. But will confirm we can have the validation to block fake email addresses Chenxing built

PR #1584, opened 2024-10-31
Mirror: changing JSON results in updates

argenisf:
Clarifying in the docs that updating JSON objects in a column, even if to add keys with null values will change the checksum and thus cause an update to the row/event.


pages/docs/tracking-methods/data-warehouse.mdx
```diff
@@ -419,6 +419,8 @@ previous section.
 ALTER TABLE <table> ADD COLUMN <column> STRING NULL;
 ```
 
+If you have a JSON column in the table/view which you map to `JSON Properties` in the import setup, as the whole JSON object for that column is used to calculate the checksum for the row, **any changes to the JSON object will result in a change to the checksum**, even if the change is to add new keys with null values. As such, consider conditional formatting when building the JSON object to only update it when a new non-null value is available for a row.
```
kurbycchua, line 422: So this looks OK for BigQuery, but now that we also have RedShift Mirror (beta) which also is implemented by taking a snapshot, would this advice also apply for Redshift?

argenisf: @kurbycchua the concept is meant to be universal. We keep track of changes to columns (values being null or not in columns), but updating the JSON blob in any warehouse (super in Redshift) will cause updates

kathmath (approved):

PR #1596, opened 2024-11-05
[Changelog] add dark mode support for changelog pages

ryancurtis1:
- updates the hero image for light and dark
- moves some styling into light and dark modes in overrides.scss


junhouse (approved):
The PR looks good, but I noticed a few things

1. Looks like we need to do some work on making subscribe modal `dark mode` compatible as well ( screenshot attached ) 
2. Also, looks like we need bottom margin for the `view more` button at the end.

Screenshot 2024-11-06 at 9 32 00 AM

Screenshot 2024-11-06 at 9 25 06 AM

PR #1600, opened 2024-11-07
add replay url method

jakewski:


pages/docs/tracking-methods/sdks/javascript.mdx
```diff
@@ -820,6 +819,15 @@ mixpanel.get_session_recording_properties()
 ```
 Returns an empty object if there is no Replay in progress.
 
+#### Get replay URL
+
+Provides a URL that links to the current replay in Mixpanel. This is useful for debugging purposes or for adding metadata to other platforms, such as support tickets.
```
tdumitrescu: I'd provide a little more context here, namely that this will return a URL to the mixpanel.com UI, which can only be accessed by an authenticated Mixpanel user with access to the project. It is not useful for programmatic access, scripting etc.

jakewski: ok how about now?

pages/docs/tracking-methods/sdks/javascript.mdx
```diff
@@ -820,6 +819,17 @@ mixpanel.get_session_recording_properties()
 ```
 Returns an empty object if there is no Replay in progress.
 
+#### Get replay URL
+
+Returns a URL linking to the current replay in the Mixpanel UI. This URL is accessible only by authenticated users with the necessary permissions for the Mixpanel project.
+
+Useful for debugging purposes or for adding metadata to other platforms, such as support tickets.
```
tdumitrescu:
```suggestion
Returns a URL linking to the current replay in the Mixpanel UI. This URL is accessible only by authenticated users with the necessary permissions for the Mixpanel project. This is useful for debugging purposes or for adding metadata to other platforms, such as support tickets.
```

pages/docs/tracking-methods/sdks/javascript.mdx
```diff
@@ -820,6 +819,17 @@ mixpanel.get_session_recording_properties()
...
+
+```javascript
+mixpanel.get_session_replay_url()
```
tdumitrescu, line 828: add sample output like we do above for `get_session_recording_properties()`?

tdumitrescu (approved):
lgtm%c

Tofufu (approved):

PR #1606, opened 2024-11-08
[ChangeLog] Add states to Watch the video button in post

junhouse:
Light
Screenshot 2024-11-08 at 3 41 48 PM

Screenshot 2024-11-08 at 3 41 51 PM

Dark
Screenshot 2024-11-08 at 3 41 35 PM
Screenshot 2024-11-08 at 3 41 38 PM
Screenshot 2024-11-08 at 3 41 43 PM

### DIFF
--- a/components/VideoButtonWithModal/VideoButtonWithModal.tsx
+++ b/components/VideoButtonWithModal/VideoButtonWithModal.tsx
@@ -12,6 +12,8 @@ type VideoButtonModalProps = {
   showThumbnail?: boolean;
 };
 
+const PLAY_BUTTON_BASE_CLASS = "nx-flex nx-items-center nx-px-2 nx-py-1 nx-text-xs nx-font-semibold nx-text-purple140 nx-shadow-sm focus-visible:nx-outline focus-visible:nx-outline-2 focus-visible:nx-outline-offset-2 focus-visible:nx-outline-purple140";
+
 export default function VideoButtonWithModal({
   title = "Video about this feature",
   showThumbnail = true,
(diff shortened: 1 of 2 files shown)


pages/overrides.scss
```diff
@@ -1522,6 +1559,35 @@ article img {
       }
     }
 
+    .thumbnailPlayButton {
```
lindseybradford, line 1562: Worth adding `1564`, `1571`, `1577` to the `colors` API?

lindseybradford (approved):

ryancurtis1 (approved):

components/VideoButtonWithModal/VideoButtonWithModal.tsx
```diff
@@ -22,8 +22,17 @@ export default function VideoButtonWithModal({
 
   // TODO: update this style and abstract it as time allows to a single button component
   // https://www.figma.com/design/8kiticjQNChvsP9y7s9SRf/Product-Releases-(Copy)?node-id=982-75355&node-type=frame&t=O7vwnwoAoOx42stw-0
+
+  let playButtonBaseClass =
```
matthathornmp: Would suggest making this a file-level const, e.g. `const PLAY_BUTTON_BASE_CLASS = "nx-flex...."`. Then you can avoid the mutable state:
```Typescript
  const playButtonStyles = showThumbnail ? `thumbnailPlayButton ${PLAY_BUTTON_BASE_CLASS}` : `playButton ${PLAY_BUTTON_BASE_CLASS};
```

pages/overrides.scss
matthathornmp: At the very least, `#FFF` is already something we can set to `colors.$white`.

matthathornmp (commented):
LGTM with a couple small comments.

pages/overrides.scss
junhouse: Updated the FFF, but for the other hex codes for now it's one-off usage of hex codes, so let me talk to designers and see if we should name them. I will follow up

PR #1611, opened 2024-11-13
[Changelog] Fix LoadMore logic

junhouse:
* Fixes an issue where `LoadMore` button appears when there is nothing else to load
* Fixes an issue where changing tab keeps the total number of items displayed, it should get reset.


lindseybradford (approved):
Nice and concise!

PR #1624, opened 2024-11-14
[ChangeLog] Different image background

junhouse:
* Image background gradient update
* Remove margin-top if no image exists above the heading

Screenshot 2024-11-15 at 11 35 58 AM
Screenshot 2024-11-15 at 11 36 02 AM

### DIFF
--- a/components/ChangelogIndex.tsx
+++ b/components/ChangelogIndex.tsx
@@ -110,7 +110,7 @@ export default function ChangelogIndex({ more = "Learn More" }) {
           <div className="changelogIndexItemBody">
             {page.frontMatter?.thumbnail && renderImage(page)}
 
-            <h3 className="changelogItemTitleWrapper">
+            <h3 className={page.frontMatter?.thumbnail && "changelogItemTitleWrapper"}>
               <Link
                 href={page.route}
                 style={{ color: "inherit", textDecoration: "none" }}
--- a/pages/overrides.scss
+++ b/pages/overrides.scss
@@ -897,6 +897,10 @@ article img {
           color: colors.$grey80;
         }
 
+        .darkImageFrame {
+          display: none;
+        }
+
         .changelogReadMoreLink {
           color: colors.$purple140;
         }
@@ -1667,6 +1671,10 @@ article img {
 
         .changelogIndexItemBody {
           color: colors.$white;
+        }
+
+        .lightImageFrame {
+          display: none;
         }
 
         .changelogReadMoreLink {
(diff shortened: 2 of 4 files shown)


components/ChangelogIndex.tsx
```diff
@@ -107,7 +107,7 @@ export default function ChangelogIndex({ more = "Learn More" }) {
           <div className="changelogIndexItemBody">
             {page.frontMatter?.thumbnail && renderImage(page)}
 
-            <h3 className="changelogItemTitleWrapper">
+            <h3 className={page.frontMatter?.thumbnail && "changelogItemTitleWrapper"}>
```
ryancurtis1, line 113: It looks like this would completely remove the title if there is no image?

ryancurtis1 (commented):
Idk if I'm looking at the wrong thing but I don't see the image frames in light or dark mode 
Screenshot 2024-11-14 at 5 13 37 PM
and I'm seeing duplicates of the images themselves

junhouse:
> Idk if I'm looking at the wrong thing but I don't see the image frames in light or dark mode Screenshot 2024-11-14 at 5 13 37 PM and I'm seeing duplicates of the images themselves

interesting, let me check.

components/ChangelogIndex.tsx
junhouse: no, this just applies `changelogItemTitleWrapper` class only when there is an `image`

tailwind.config.js
```diff
@@ -45,6 +45,10 @@ module.exports = {
         grey50: "#9b9b9b",
         grey30: "#c0c0c0",
         grey20: "#eae7e7",
```
junhouse, line 47: I've been fighting w/ tailwind here, had to extract out colors to make it work.

pages/overrides.scss
```diff
@@ -897,6 +897,10 @@ article img {
           color: colors.$grey80;
         }
 
+        .darkImageFrame {
```
ryancurtis1: I think the issue with duplicate images on the individual post pages is because these css overrides are nested in the `changelogIndexItem` but that component doesn't exist on the post page

junhouse: hm, what are you referring to w/ duplicate images?

ryancurtis1: There are two copies of the header image on every post page for me

Screenshot 2024-11-15 at 2 16 42 PM

junhouse: Can you make sure to fetch all, and pull? It looks fine on the preview @ryancurtis1

ryancurtis1: I tested both locally and on the preview link and see the issue in both places 🤔

ryancurtis1 (approved):

PR #1627, opened 2024-11-19
[changelog] show video in the feed if no image available

ryancurtis1:
Screenshot 2024-11-19 at 1 34 09 PM

### DIFF
--- a/components/ChangelogIndex.tsx
+++ b/components/ChangelogIndex.tsx
@@ -20,6 +20,45 @@ const renderImage = (page) => {
   );
 };
 
+const renderVideo = (page) => {
+  const videoURL = page.frontMatter.video;
+  let embedURL;
+
+  if (videoURL.includes("youtube.com") || videoURL.includes("youtu.be")) {
+    const videoId = videoURL.split("v=")[1]
+      ? videoURL.split("v=")[1].split("&")[0]
+      : videoURL.split("/").pop();
+    embedURL = `https://www.youtube.com/embed/${videoId}`;
+  } else if (videoURL.includes("loom.com")) {
+    const videoId = videoURL.split("/").pop();
+    embedURL = `https://www.loom.com/embed/${videoId}?hideEmbedTopBar=true`;
+  }
+
+  return (
+    <iframe
+      src={embedURL}
+      style={{
+        width: "100%",
+        aspectRatio: 16 / 9,
+        height: "auto",
+        borderRadius: "16px",
+        marginBottom: "16px",
+      }}
+      allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture"
+      allowFullScreen
+      title="Video"
+    ></iframe>
+  );
+};
+
+const renderMedia = (page) => {
+  if (page.frontMatter?.thumbnail) {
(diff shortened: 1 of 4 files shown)


pmackle (approved):

components/ChangelogIndex.tsx
```diff
@@ -20,6 +20,45 @@ const renderImage = (page) => {
...
+
+  if (videoURL.includes("youtube.com") || videoURL.includes("youtu.be")) {
+    const videoId = videoURL.split("v=")[1]
+      ? videoURL.split("v=")[1].split("&")[0]
+      : videoURL.split("/").pop();
+    embedURL = `https://www.youtube.com/embed/${videoId}`;
+  } else if (videoURL.includes("loom.com")) {
+    const videoId = videoURL.split("/").pop();
+    embedURL = `https://www.loom.com/embed/${videoId}?hideEmbedTopBar=true`;
+  }
```
matthathornmp, line 35: Would move this block to a helper function, since its responsibility is somewhat distinct from the rendering of the video. That would also allow you to use immutable state (`const embedURL`) instead of the mutable state you're using now (`let embedUrl`).

components/ChangelogIndex.tsx
```diff
@@ -20,6 +20,45 @@ const renderImage = (page) => {
...
+    <iframe
+      src={embedURL}
+      style={{
+        width: "100%",
+        aspectRatio: 16 / 9,
+        height: "auto",
+        borderRadius: "16px",
+        marginBottom: "16px",
+      }}
+      allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture"
```
matthathornmp, line 47: Why does this frame need access to the gyroscope and accelerometer?

components/ChangelogIndex.tsx
```diff
@@ -20,6 +20,45 @@ const renderImage = (page) => {
...
+    ></iframe>
+  );
+};
+
+const renderMedia = (page) => {
+  if (page.frontMatter?.thumbnail) {
+    return renderImage(page);
+  } else if (page.frontMatter?.video) {
+    return renderVideo(page);
+  }
```
matthathornmp, line 59: You should strive to avoid dangling else ifs. Sometimes they're needed, but in those cases, a comment saying why you aren't ending the block with an `else` helps to clarify your intent. In this case, though, I would suggest you type the parameters and return value, which would help you see that you're returning a JSX element and therefore that the else block (you don't need to literally write the else block, a return on this line is fine) should return null.

components/ChangelogIndex.tsx
ryancurtis1: No idea... I just copied the code back in from what is in prod for the video component. Probably safe to remove these 

original commit that added the gyroscope and accelerometer: https://github.com/mixpanel/docs/commit/2d7a6f88118577411cdd173f51abe30b7499044c

matthathornmp: Okay, let's leave it for now then. If we can figure out why it's there and that it's safe to remove it, we can go ahead. Until then, I don't think it's much of a threat to allow youtube or loom access to device movements.

PR #1628, opened 2024-11-19
[changelog] address comments and some small style fixes

ryancurtis1:
- addresses comments from https://github.com/mixpanel/docs/pull/1627
- updates margin for image frame on smaller screens
- prevent the thumbnail for videos from playing the video on post pages and makes whole button clickable to open video in a modal
- adjust width of hero image

### DIFF
--- a/pages/overrides.scss
+++ b/pages/overrides.scss
@@ -215,6 +215,7 @@ $border-radius-lg: 100px;
   font-style: normal;
   font-weight: 500;
   line-height: 24px; /* 150% */
+  position: unset;
 }
 
 .changelogMainContainerWrapper {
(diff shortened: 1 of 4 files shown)


pages/overrides.scss
```diff
@@ -1582,9 +1583,12 @@ article img {
       }
     }
 
-    .thumbnailPlayButton {
+    .thumbnailPlayButtonWrapper {
```
matthathornmp, line 1586: Is it intentional that this definition is separate from the other `.thumbnailPlayButtonWrapper` definitions above?

matthathornmp (approved):
LGTM.

pages/overrides.scss
ryancurtis1: yes, one for light mode and one for dark mode

PR #1643, opened 2024-11-22
Update concepts.mdx

danielionashku-mixpanel:
The Data_Model_Overview.png image seems to imply that users can setup a lookup table based on a group profile property. However, further down in the doc we clarify that this is not possible: 

Note: Currently, group profile properties cannot be lookup table keys.

I'm not sure if this image needs to be updated, replaced, or if I'm just misunderstanding.


myronkaifung:
Updated image to remove Lookup Table <> Group Profile since it cannot be mapped

PR #1715, opened 2025-01-10
Pinned Filters ChangeLog

jennabono-mixpanel:


catliaw:
Hey Jenna! I think the screenshot image should be updated because that is the old Pinned Filters UI. It has changed already. I think maybe something like this showing where the filter settings are would be maybe more helpful.
Screenshot 2025-01-15 at 2 25 13 PM
But make sure to use a screenshot from the demo. Not sure who owns the Pinned Filters Demo board and which project it's in.

catliaw:
Okay one more nitpicky note. On the screenshot for the UTM Campaign filter, it's showing "Any", but the selected setting is "limit filter values to selection above". Can you Apply the setting and then open the filter settings to take a screenshot? So the screenshot shows what the board filter will look like when the setting is applied. I want to distinguish from the other "Start from none" setting which will show "Any".

catliaw (approved):
Thank you for updating the screenshot :D! Looks good! Approved

PR #1740, opened 2025-01-29
redistribute advanced

myronkaifung:


myronkaifung:
Redistributed content under Boards>Advanced

- Created a page for sharing and permissions
- Created a page for moving boards
- Content on board-wide filters/date range, subscription, and TV mode moved to Boards main page

PR #1746, opened 2025-01-31
Update data-clean-up.mdx

djmixpanel:


Tofufu (commented):
Rather than "Distinct ID Deletion API", should we instead rephrase to GDPR/CCPA API (as that is the name of the API)?

PR #1757, opened 2025-02-04
Update Group Analytics Docs

nsquare92:
Added B2B Company Analytics setup instructions and FAQs on how this is An advanced offering within Group Analytics


myronkaifung (approved):

myronkaifung:
Approving for now, but we should revisit the content here for completeness. It is not clear from the docs how setting up a Company key is different from setting up a Group Key. As well, highlighting these features more appropriately in the page instead of as a footnote buried in the FAQ

PR #1762, opened 2025-02-06
docs-feedback-patch-20250205

myronkaifung:
- Add clarification about when 50k csv is supported
- Add clarification about Lexicon UI vs Lexicon Schema API discrepancy
- Add clarification about Service Account role required for Pipelines API in docs and dev docs
- Removed reference to last touch UTM tag custom setup since this is the default behavior for JS SDK
- Make UTM section of JS SDK more pleasing to the eyes
- Emphasize recommendation to disable UTM persistence, most compatible with attribution models


myronkaifung:
- Add clarification about when 50k csv is supported
- Add clarification about Lexicon UI vs Lexicon Schema API discrepancy
- Add clarification about Service Account role required for Pipelines API in docs and dev docs
- removed reference to last touch UTM tag custom setup since this is the default behavior for JS SDK
- make UTM section of JS SDK more pleasing to the eyes
- emphasize recommendation to disable UTM persistence, most compatible with attribution models

PR #1779, opened 2025-02-24
Periodically clean up stale issues and pull requests

austinpray-mixpanel:
Implements https://github.com/actions/stale to clean up PRs and issues that have been sitting for over 180 days without any sort of activity. Prompts users to rebase their PRs or bump their issues with interest.

This is the same action we use to keep our monorepo tidy


matthathornmp (approved):
LGTM.

PR #1799, opened 2025-03-12
Add Shopify Guide to Docs

lindseybradford:
This PR adds a Mixpanel + Shopify Integration Guide to Docs and updates the Integration resources/link lists

Screenshot 2025-03-12 at 4 38 30 PM

While I was digging into the docs content, I realized that we also should move the [current Nextjs guide](https://docs.mixpanel.com/docs/quickstart/install-mixpanel/nextjs) from the Quickstart section into the Tracking > Integrations section with the rest of the guides. I made that update and _added the complete list of integrations to the Quick Start page_, replacing the single nextjs card:
Screenshot 2025-03-12 at 4 42 04 PM

…with a larger list of resources:
Screenshot 2025-03-12 at 4 43 18 PM

I updated the nextjs paths with this change, too. 👍🏻 

I simplified the existing references to shopify integrations on the CMS page ([current version](https://docs.mixpanel.com/docs/tracking-methods/integrations/cms-ecommerce#shopify)) since that info is outdated and a lot simpler…and, now covered in the new guide! 🎉 

I checked everything in my own shopify dev store to make sure all the code examples work and were tracked in my a Mixpanel test project (which you can see in the screenshots). 

Question for @junhouse - Is it OK I moved your next guide? Are there any strings attached to this change like tracking I should go and update as well? And speaking of tracking, I saw what looked like a dynamic tracking function in `_app.ts`. Did you add anything more detailed for the Next file/would you recommend anything more here?

Lastly, I've got these set of copy updates out for review by Isha so we can sanity check. 

Thanks!


lindseybradford:
PS: I _dislike_ the out of the box spacing we get with nextra between text and images. So I'm going to push on that a little more after this review.

abraibrai (approved):

pages/docs/tracking-methods/integrations/shopify.mdx
```diff
@@ -0,0 +1,150 @@
...
+            productPrice: event.data.productVariant.price.amount
+        })
+    });
+    ```
+
+    Putting this recipe to work, here’s an example of how to track `product_viewed` events:
+    
+    ```javascript
+    // Subscribe to Shopify Pixel Event API
+    analytics.subscribe('EVENT_NAME', (event) => {
```
junhouse, line 87: Update `EVENT_NAME` to `product_viewed`.

junhouse (commented):
@lindseybradford -  This is amazing! I can't wait to set up the part 2 (pixel based approach) to my store - I am going to reference this guide. 

A few things, 

* We should consider moving Next.js and Shopify guide back to Framework guides section.  Integration means they already have the tracking data somewhere (e.g. Segment), and they want to send the data TO mixpanel, whereas Shopify guide is about HOW to implement mixpanel. 

* I would either delete or obfuscate your project token.

Overall, thanks for this work!

lindseybradford:
> @lindseybradford - This is amazing! I can't wait to set up the part 2 (pixel based approach) to my store - I am going to reference this guide.
> 
> A few things,
> 
> * We should consider moving Next.js and Shopify guide back to Framework guides section.  Integration means they already have the tracking data somewhere (e.g. Segment), and they want to send the data TO mixpanel, whereas Shopify guide is about HOW to implement mixpanel.
> * I would either delete or obfuscate your project token.
> 
> Overall, thanks for this work!

Re: pixel setup - Thanks! LMK how it goes and feel free to pass along any feedback you have when you run through it in you store.

Re: token - Abby noted that some thing and had me change the heading to Integration / Framework Guides:
Screenshot 2025-03-13 at 11 23 53 AM

Do you feel that covers your concern noted above?

Re: obfuscate token - good point, will do!

PR #1820, opened 2025-04-01
Update warehouse-connectors.mdx

ishamehramixpanel:


pages/docs/tracking-methods/warehouse-connectors.mdx
```diff
@@ -651,18 +651,44 @@ We recommend that customer filter such events and refresh such events when they
...
+| **Historical events** | Not billed | If you send Event A with properties a, b, c, d as a historical event (from a past month), it will not count toward billing |
+| **Event table updates** | Billed for the month the update happens | If you already have Event A in your database and add property e or update property b's value, each affected row counts as a billable event |
+| **Event table column additions** | Billed for non-NULL values | Adding a column with NULL values does not count as an update across all supported warehouses, each affected row counts as a billable event otherwise |
+| **Event table deletes** | Billed when the delete occurs | Deleting events or properties will generate billable events for each affected row |
+| **User table standard operations** | Not billed | In standard mode, user profile inserts, updates, and deletes are not billed |
+| **User table history mode** | Billed for updates and deletes | With history sync enabled, each profile update or delete generates a billable event, including the initial historical import |
```
mherrman: my understanding is that history/mirror mode user profile syncs will be billed exactly like a mirror event sync so from a historical import perspective, we would only bill for profile history where the timestamp (start_time) is in the current month, similar to how events work. Can maybe double check with Sowmya/Deepak to confirm if this was something that came from conversation with them

pages/docs/tracking-methods/warehouse-connectors.mdx
```diff
@@ -647,16 +647,69 @@ We recommend that customer filter such events and refresh such events when they
...
-On an [MTU billing plan](/docs/pricing/legacy-mtu-billing) updates and deletes from Mirror do not impact [MTU calculations](/docs/pricing/legacy-mtu-billing#mtu-calculation), however updates and deletes are counted as events for the purposes of the [MTU guardrail](/docs/pricing/legacy-mtu-billing#mtu-guardrail) computation.
+**Billing for Event Syncs:**
+
+| **Event Operation**                        | **Sync Type** | **Supported** | **Billed** |
+|-------------------------------------------|---------------|---------------|-------------|
+| **New events (current calendar month)**  | One time      | ✅            | ✅          |
+|                                           | Append        | ✅            | ✅          |
+|                                           | Mirror        | ✅            | ✅          |
+| **Historical events (backfills)**         | One time      | ✅            | ❌          |
+|                                           | Append        | ❌            | ❌          |
```
mherrman, line 665: curious as to the logic here. As i understand it there's no reason you could not import historic data with an append sync as long as you have different event time and inset time columns right?

pages/docs/tracking-methods/warehouse-connectors.mdx
```diff
@@ -647,16 +647,69 @@ We recommend that customer filter such events and refresh such events when they
...
+
+This will count as an update event for each row in the sync.
+
+If your warehouse workflow **drops and recreates tables**, Mirror will treat this as a full delete and re-insert, with each row counted as a billable event. Mirror looks at the data already ingested into Mixpanel and uses warehouse-specific logic to identify changes made within your warehouse platform. In the case of a full table delete, Mixpanel will register each deleted row as a billable event, and the same applies when the table is recreated.
+
+If instead your workflow updates existing tables—by appending, updating, or deleting specific rows or columns—only the affected rows will be billed, as shown in the table above.
+
+**Note on Deletes**: If you delete an event using the Mixpanel UI instead of Mirror, it will not be counted as a billable event.
+
+**Note on Backfills**: If you already have an event in your Mixpanel project, and you sync in historical events for a past month with the same properties, it will not count towards billing. Otherwise, billing will occur at the time of the event update.
```
mherrman: is this saying if you send a duplicate you won't be billed for it essentially?

dneralla: Can reword deletes and I think we can remove backfills. Anything in historical is not billed unless updated from mirror anyway right. 

**Note on Deletes**: If you delete events using [Mixpanel UI](https://docs.mixpanel.com/docs/data-governance/data-clean-up) , those deletes are not counted towards billable events. Only Deletes coming from warehouse using mirror are billable.

pages/docs/tracking-methods/warehouse-connectors.mdx
dneralla: @mherrman true but in an ongoing way, once u create a sync and want to backfill whole data again ..its not effectively possible.

pages/docs/tracking-methods/warehouse-connectors.mdx
dneralla: **Note on backfills**:  You can also backfill using `Append` mode if you create a new sync. But for a ongoing sync you cannot backfill for older days with in the existing sync once the `insert_time` has moved past.

pages/docs/tracking-methods/warehouse-connectors.mdx
dneralla: I added some wording in the other comment on how we can add it.

dneralla (approved):

PR #1833, opened 2025-04-09
[Security] Update security headers

tiffanyqi:
https://mixpanel.atlassian.net/browse/SECOPS-770

Main thing here is whether it not breaks Kapa AI. It shouldn't, but I will rollback if it does.


mherrman (approved):
im no expert here by any means but seems like this sets what the task is asking for

PR #1850, opened 2025-04-23
data governance - update asset

spjtls9:


dongjae93 (approved):
LGTM. One question: The original looks like it had transparent background and the new one doesn't. Is that intentional?

spjtls9:
> LGTM. One question: The original looks like it had transparent background and the new one doesn't. Is that intentional?

Yes -- I think it's the transparent background that made it look weird.

PR #1862, opened 2025-05-06
[Docs] remove max width on docs

ryancurtis1:
slack thread: https://mixpanel.slack.com/archives/C050U6CQRLZ/p1745830340536249

screenshot of changes:
Screenshot 2025-05-06 at 3 51 32 PM

### DIFF
--- a/pages/overrides.scss
+++ b/pages/overrides.scss
@@ -467,6 +467,10 @@ $border-radius-lg: 100px;
   opacity: 1 !important;
 }
 
+._max-w-\[90rem\] {
+  max-width: 100% !important;
+}
+
 // Add image frame styling to prevent layout-reflow and match custom layout image styling (changelog) with nextra layout
 
 article {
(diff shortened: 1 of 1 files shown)


pages/overrides.scss
```diff
@@ -467,6 +467,10 @@ $border-radius-lg: 100px;
   opacity: 1 !important;
 }
 
+._max-w-\[90rem\] {
```
lindseybradford, line 470: Double clicking on why the utility name doesn't match the rule: Is it possible to remove this class from where it's set or is this part of some generated markup somewhere?

ryancurtis1: The reason the rule and classname don't perfectly match is because the `[]` are protected characters so the `\` is needed as an escape character

It looks like this class is being applied as part of the generated code from nextra and I am not able to edit that part, can only override the rule here. Sorta hacky, I know, but nextra is not entirely customizable

lindseybradford: For sure, I understand the protected characters (super nifty and common utility naming convention!), but I see--it's nextra generated. Booo! Have to do what you have to do.

lindseybradford (approved):

PR #1870, opened 2025-05-14
Remove from object limits to contact support to request higher lmt

catliaw:
Removing that customers with a paid plan can write in the request > 20 object limits. We are only currently only entertaining this for customers who have a plan with $100k+ ARR. "Paid plan" is too broad. Also most customers with $100k+ ARR should have an account team to facilitate this request internally.


catliaw:
Err, I need to add more edits, so will create another PR.

PR #1871, opened 2025-05-14
Remove mentions to contact support to request higher object lmt 

catliaw:
Removing that customers with a paid plan can write in the request > 20 object limits. We are only currently only entertaining this for customers who have a plan with $100k+ ARR. "Paid plan" is too broad. Also most customers with $100k+ ARR should have an account team to facilitate this request internally.


myronkaifung:
@nsquare92 to review

Tofufu (approved):

PR #1915, opened 2025-06-24
[incr pipelines] docs updates

howardchen11:


howardchen11:
from Mikey: Couple questions I have from reading through:
- whats happens to customers who have a daily and hourly event pipeline set up when we try to migrate them?
- is there a way for us to know if the customer's pipeline is using incremental behavior or not when debugging issues (specifically thinking of during the transition period we need to know what to expect)

dneralla (approved):

santigracia:
LGTM!

PR #1929, opened 2025-07-03
pipelines docs update: redshift & schema column

howardchen11:
https://mixpanel.slack.com/archives/C9QJ37WC8/p1751446713778829

### DIFF
--- a/pages/docs/data-pipelines/integrations/redshift-spectrum.mdx
+++ b/pages/docs/data-pipelines/integrations/redshift-spectrum.mdx
@@ -256,7 +256,7 @@ Follow these steps to grant the `CREATE` privilege:
 
   ```sql
   -- Check if the IAMR user has been created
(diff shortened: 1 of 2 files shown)


pages/docs/data-pipelines/integrations/redshift-spectrum.mdx
```diff
@@ -256,7 +256,7 @@ Follow these steps to grant the `CREATE` privilege:
 
   ```sql
   -- Check if the IAMR user has been created
-  SELECT * FROM pg_user WHERE username LIKE 'IAMR:<mixpanel-role-name>';
+  SELECT * FROM pg_user WHERE usename LIKE 'IAMR:<mixpanel-role-name>';
```
howardchen11, line 259: bug I found when testing

mherrman (approved):

PR #1949, opened 2025-07-15
add canonical url

tiffanyqi:
canonical URLs are how search engines determine what the "main" or "official" link is. while docs largely doesn't have a problem with this since we don't do too many displays of the same page with different urls, ahrefs still called a few of them out, so I figured I'd add it to docs as well.

the canonical URL will be:

https://docs.mixpanel.com/docs/quickstart/install-mixpanel when you go to https://docs.mixpanel.com/docs/quickstart/install-mixpanel#overview , for example

### DIFF
--- a/theme.config.tsx
+++ b/theme.config.tsx
@@ -1,6 +1,6 @@
-import { FC, ReactNode } from "react";
 import { useRouter } from "next/router";
 import { DocsThemeConfig, useConfig } from "nextra-theme-docs";
+
 import Search from "./components/Search/Search";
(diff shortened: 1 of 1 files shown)


theme.config.tsx
```diff
@@ -15,14 +15,15 @@ const config: DocsThemeConfig = {
   },
   docsRepositoryBase: "https://github.com/mixpanel/docs/tree/main",
   head() {
-    const { asPath } = useRouter();
+    const { pathname } = useRouter();
```
tiffanyqi, line 18: pathname strips the search params, which is why this is a better canonical url

danszemix (approved):

PR #2010, opened 2025-08-14
Myronkaifung redistribute featuresadvanced

myronkaifung:


pages/docs/data-structure/group-analytics.mdx
```diff
@@ -98,6 +98,37 @@ Company Analytics is an advanced offering ‘within’ group analytics tailored
 
 Company Analytics enables you to understand such account health and activation metrics, which are derived from the underlying user activity.
 
+To use these features, you must [designate the group key as a company key when setting up the group](/docs/data-structure/group-analytics#setup-b2b-company-key).
+
+#### Company Profiles
+
+This page gives you a view into all your company health - including key metrics, and company profile properties (i.e company attributes)
```
Tofufu: can we add periods to the sentences that were added here that aren't bullet points?

pages/docs/data-structure/group-analytics.mdx
```diff
@@ -98,6 +98,37 @@ Company Analytics is an advanced offering ‘within’ group analytics tailored
...
+You can view the definition of each metric by clicking on the metric card, which will open the underlying Mixpanel report 
+
+1. To access this page - click on Users/ Companies → Company Profile
+2. This is only available if you have [set up a B2B Company Key](/docs/data-structure/group-analytics#setup-b2b-company-key), an option available in the Group Analytics package.
+
+![image](/B2B_Company_Profiles.webp) 
+
+#### Activation Metrics 
+
+SaaS companies often have use cases like the one below, where you’re trying to segment companies based on how many and the quality of users. (Examples - what are the number of trial accounts that have > 2 active users?)  The hypothesis is that these will convert faster. What you want to do here is to be able to break down account activity by the number of active users
```
Tofufu: maybe can change or rephrase to something like the below: 
"...segment companies based on the number and quality of users. 

For example, a question you may have is 'What is the number of trial accounts that have more than two active users?'. The hypothesis is that these accounts will convert faster.

We now have a new computed property (in both breakdown and filter) - “Number of users who did…”, available in our group analytics package, that enables you to answer this question. 

It allows you to break down account activity by the number of active users.
".

myronkaifung: Done

Tofufu (approved):

PR #2025, opened 2025-08-22
Add Backup Host and Tracking via proxy to Android docs

santigracia:
- v8.2.2 has been released, adding the ability to set a backup host. 
- docs were also missing the section for tracking via proxy


argenisf (approved):
thanks Santi!

PR #2034, opened 2025-08-25
Remove markdown-link-check

myronkaifung:


tiffanyqi (commented):
fine by me, would you like to handle the dependency deletion in package.json as well?
the process would be to delete the package, and then run `npm install` within the docs repo in your terminal, and then push the output of both package.json and package-lock.json.

myronkaifung:
> fine by me, would you like to handle the dependency deletion in package.json as well? the process would be to delete the package, and then run `npm install` within the docs repo in your terminal, and then push the output of both package.json and package-lock.json.

Let me give that a try, thanks!

myronkaifung:
@tiffanyqi I followed the steps, but it doesn't look like there were any dependencies removed from package.json when I ran npm install locally. Could you take a look for me please?

tiffanyqi:
my mistake! you're totally right, looks like it's only referencing a github action and it wasn't installed as an npm package.

tiffanyqi (approved):

PR #2046, opened 2025-08-29
python-sdk-geolocation_update

IlhanMP:
I have updated the Python SDK to reflect the source code. The IP address is dropped, so it should be explicitly set in event and profile updates.


IlhanMP:
[Here is a link to the source code.](https://github.com/mixpanel/mixpanel-python/blob/b53d31584acf2976d106994960f27264efa534ec/mixpanel/__init__.py#L605)

argenisf (approved):
@IlhanMP looks good to me :)

PR #2084, opened 2025-09-15
Upgrade redocly

tiffanyqi:
the beta is deprecated: https://www.npmjs.com/package/@redocly/openapi-cli?activeTab=versions
in favor https://www.npmjs.com/package/@redocly/cli/v/1.34.5
I picked 1.34.5 because while it isn't the latest version, v2 is on node 20.19, and we are on 20.18.

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
     "sharp": "0.34.3"
   },
   "devDependencies": {
-    "@redocly/openapi-cli": "1.0.0-beta.95",
+    "@redocly/cli": "1.34.5",
     "@types/mixpanel": "2.14.9",
     "@types/node": "24.3.0",
     "autoprefixer": "10.4.21",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.0.1",
   "description": "Mixpanel's Official Documentation",
   "scripts": {
-    "api:build": "openapi --config=./openapi/openapi.config.yaml --output ./openapi/out/ bundle",
+    "api:build": "redocly bundle --config=./openapi/openapi.config.yaml",
     "api:lint": "openapi --config=./openapi/openapi.config.yaml --format=stylish lint",
```
junhouse, line 7: Do we need to specify the output bundle path here as well?

tiffanyqi: we do not. the docs have changed, now if we supply --output, we need to specify every single yaml file. instead, i've added an output field directly in the config file which does the same thing. here are the docs: https://redocly.com/docs/cli/commands/bundle#options

junhouse (approved):

PR #2091, opened 2025-09-18
Fixed instructions for NextJS App Router integration

tomredman:
The current docs say to use `useEffect()` in the root layout, but that can't be done as the root layout is a special case, and is always server component, regardless of the "use client" directive. You cannot use `useEffect()` in a server component.


hans-lizihan:
closing as duplicate of https://github.com/mixpanel/docs/pull/1953

thanks for the contribution!

PR #2093, opened 2025-09-19
Improving docs around Historic People and Group property values

argenisf:
- Added more detail to the warehouse import page on structure and requirements
- Expanded within data structure on the implementation (events associated and the attribution mechanism)


nsquare92 (commented):
Content looks good to me -  But a couple things

1. i'd rather mention the billing details in - https://docs.mixpanel.com/docs/tracking-methods/warehouse-connectors#billing-faq. Basically in just data structure would talk about historical profiles and how they're modelled; but keep all billing deets in the WH section
2.  Can we talk. about custom props leveraging SCD, and say these custom props always use value as of time? 
3. If you have the bandwidth @argenisf - could i request to make the following doc edits too pls? https://www.notion.so/mxpnl/Profile-History-Docs-Improvements-1f2e0ba925628042b612eaa326e82eec

nsquare92 (approved):
Looks good to me!

PR #2094, opened 2025-09-19
Update reports.mdx

wangweiyi722:
Cleaned up the first portion of the reports page, particularly the metrics,measurements and behaviors portion.


myronkaifung (approved):
Thanks @wangweiyi722

PR #2124, opened 2025-09-30
Fix openapi

tiffanyqi:
Docs: https://github.com/readmeio/rdme/blob/next/documentation/commands/openapi.md#rdme-openapi-upload-spec

Currently, the openapi script must be updated to correspond to readme refactored.

### DIFF
--- a/.github/workflows/rdme-openapi.yml
+++ b/.github/workflows/rdme-openapi.yml
@@ -1,7 +1,8 @@
-name: Publish OpenAPI specs to ReadMe
+name: Publish OpenAPI specs to ReadMe Production
 
 on:
   push:
+    branches: [ main ]
     paths:
       - 'openapi/**'
       - '.github/actions/**'
--- a/.github/workflows/rdme-openapi.yml
+++ b/.github/workflows/rdme-openapi.yml
@@ -2,7 +2,6 @@ name: Publish OpenAPI specs to ReadMe
 
 on:
   push:
-    branches: [ main ]
     paths:
       - 'openapi/**'
       - '.github/actions/**'
--- a/.github/workflows/rdme-openapi.yml
+++ b/.github/workflows/rdme-openapi.yml
@@ -3,6 +3,10 @@ name: Publish OpenAPI specs to ReadMe
 on:
   push:
     branches: [ main ]
+    paths:
+      - 'openapi/**'
+      - '.github/actions/**'
+      - '.github/workflows/rdme-openapi.yml'
 
 jobs:
   rdme-openapi:
@@ -17,4 +21,4 @@ jobs:
     - run: npm run api:publish
       env:
         README_API_KEY: ${{ secrets.README_API_KEY }}
-        README_VERSION: '3.26'
+        README_VERSION: '3.27'
--- a/openapi/openapi.config.yaml
+++ b/openapi/openapi.config.yaml
@@ -1,34 +1,34 @@
 apis:
   annotations:
     root: ./src/annotations.openapi.yaml
-    output: ./out/annotations.openapi.yaml
+    output: ./out/annotations-api.json
   data-pipelines:
     root: ./src/data-pipelines.openapi.yaml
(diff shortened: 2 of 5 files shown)


openapi/openapi.config.yaml
```diff
@@ -1,34 +1,34 @@
 apis:
   annotations:
     root: ./src/annotations.openapi.yaml
-    output: ./out/annotations.openapi.yaml
+    output: ./out/annotations-api.json
```
tiffanyqi, line 4: these names correspond to the ones in our api reference

.github/workflows/rdme-openapi.yml
```diff
@@ -2,7 +2,10 @@ name: Publish OpenAPI specs to ReadMe
 
 on:
   push:
-    branches: [ main ]
```
tiffanyqi, line 5: i'll put this back after review

openapi/publish.js
```diff
@@ -47,25 +44,30 @@ async function updateSpecs() {
...
 
     // validate and publish spec
-    console.log(`Updating ${spec.info.title} (${specFile}, ID ${specId})`);
-    await execAndLog('npx', ['rdme', 'openapi:validate', fullPath]);
+    console.log(`Updating ${spec.info.title} (${specFile})`);
+    await execAndLog('npx', ['rdme@10.5.1', 'openapi:validate', fullPath]);
+    // publish the json version
     await execAndLog(
-      'npx', ['rdme', 'openapi', fullPath, `--id=${specId}`, `--key=${README_API_KEY}`]
+      'npx', ['rdme@10.5.1', 'openapi', 'upload', `${fullPath}`, `--key=${README_API_KEY}`, `--slug=${specFile}`, `--confirm-overwrite`],
```
gabrielvigil-mixpanel: On the rdme docs it looks like this command expects some kind of branch value/version. Since this is used for both the production github action on merge to main and the staging action how does this differentiate between the two? I saw previously there was an ID value that was replaced by the slug, does this value also point to branch/version?

tiffanyqi: yes. by default, it will post directly to the default version, which in our case is version 3.27 (you might see that tossed around). The github action production script (rdme-openapi) will post this to version 3.27, while within rdme-staging (puts content from both openapi and reference within our dev docs in developer.mixpanel.com to a specific staging version). You might have seen that I commented out the main branch for testing purposes, but I plan to bring it back when I get a successful review; it's more so that you can see what it's supposed to be doing.

gabrielvigil-mixpanel: Ok yeah I see the README_VERSION at 3.27 in the rdme-openapi script and README_VERSION set to a value with the github event number that is then passed to this script in rdme-staging.yml. I guess what I'm confused about is it doesn't look like the `README_VERSION` value is used in the command directly so I would expect [based on the documentation](https://github.com/readmeio/rdme/blob/next/documentation/commands/openapi.md#rdme-openapi-upload-spec) that the branch would default to `stable`. Does the command take directly from the env var or maybe there is a config I'm missing to be used in this command? The previous iteration didn't have branch/version defined either so I assume I'm missing something here.

openapi/publish.js
```diff
@@ -47,25 +44,30 @@ async function updateSpecs() {
...
-    .filter((fn) => fn.endsWith(`.openapi.yaml`));
+    .filter((fn) => fn.endsWith(`.json`));
+
+  if (!remoteSpecMetas) {
+    console.error(`!!! No remote specs found, please double check the API`);
+    process.exit(1);
+  }
 
-  for (specFile of filenames) {
+  for (specFile of [filenames[0]]) {
```
tiffanyqi: TODO remove this too, this was only for testing

openapi/publish.js
tiffanyqi: oh! I think you're totally right. I'm not super sure about the previous iteration, perhaps the ids are different across branches? I will modify this

gabrielvigil-mixpanel (approved):
lgtm

PR #2169, opened 2025-10-14
Updating pipelines documentation

xyn1nja:
- Added FAQ on what happens to data older than 10 days (late-arriving data)
- Added FAQ on whether it's possible to change the destination of an existing pipeline
- Added FAQ on whether it's possible to customise subfolders when exporting to a bucket
- Added FAQ on whether it's possible to only export a subset of data using JSON pipelines
- Added a section on supported GCP regions for BigQuery


argenisf:
hey @xyn1nja, on the section "what happens to data older than 10 days", is it worth having? nearly all pipelines now (except some enterprise ones) have been migrated to incremental. The 10 day reference will soon not make much sense as a cutoff. It'd be better to rephrame it to late arriving data in general and make the distinction between pipelines with sync and without it (which would only be raw + schematized pipelines created without sync)

argenisf (changes requested):
requesting to review section on late arriving data

argenisf (approved):

PR #2181, opened 2025-10-22
add Java Feature Flags SDK

msiebert:


efahk:
image
On the main page, https://docs-git-msiebert-java-ff-sdk-mixpanel.vercel.app/docs/featureflags, can you add the link for Java like the others and remove it from the 'coming soon'?

efahk:
image

- Definitely don't block on this, but maybe for the GA version, i'd recommend we update the startPolling to block on the 'first' flag definitions fetch, similar to the python SDK, so the customer doesn't have to wait for readiness in a timer loop

efahk (approved):
Couple of minor comments - lgtm

msiebert:
> image On the main page, https://docs-git-msiebert-java-ff-sdk-mixpanel.vercel.app/docs/featureflags, can you add the link for Java like the others and remove it from the 'coming soon'?

@copilot please implement this change

msiebert:
> image
> * Definitely don't block on this, but maybe for the GA version, i'd recommend we update the startPolling to block on the 'first' flag definitions fetch, similar to the python SDK, so the customer doesn't have to wait for readiness in a timer loop

Sorry, I'm not sure what you mean? I get removing the block with the `Thread.sleep` but what do you mean about the start polling part?

msiebert:
> > image
> > * Definitely don't block on this, but maybe for the GA version, i'd recommend we update the startPolling to block on the 'first' flag definitions fetch, similar to the python SDK, so the customer doesn't have to wait for readiness in a timer loop
> 
> Sorry, I'm not sure what you mean? I get removing the block with the `Thread.sleep` but what do you mean about the start polling part?

@copilot we should just remove all the references to waiting for flags to be ready, since we fetch them automatically when starting polling. No need to add a comment describing that behavior, just remove the code blocks with Thread.sleep

msiebert:
@copilot since the SDK is currently in beta and we are using a `-SNAPSHOT` version, please include in the docs that the users need to set up pulling in snapshot versions. You can link to the docs for how to do this https://central.sonatype.org/publish/publish-portal-snapshots/#publishing-via-other-methods

msiebert:
@copilot add "Sonatype's" and "Sonatype" to the cspell library

PR #2182, opened 2025-10-23
Xyninja/pipelines 5

xyn1nja:
Added the following SQL queries to a new page in the docs:
- Count of total events per day in Pacific timezone, deduplicated by `event_name`, `time`, `distinct_id`, and `insert_id`
- Count of unique users per day (joining the events table with the identity mapping table) in Pacific timezone
- Top 20 events by volume
- Querying duplicate events (by `event_name`, `time`, `distinct_id`, and `insert_id`)


pages/docs/data-pipelines/common-sql-queries.mdx
```diff
@@ -0,0 +1,101 @@
+# Common SQL Queries
+
+### Count of total events per day in Pacific timezone, deduplicated by `event_name`, `time`, `distinct_id`, and `insert_id`
```
argenisf: Hey @xyn1nja could we rework the titles to something more concise and give a brief explanation of the use case? Here's an example of the first 2 queries:

image

pages/docs/data-pipelines/common-sql-queries.mdx
```diff
@@ -0,0 +1,101 @@
...
+ AND DATE(time, 'America/Los_Angeles') < '2025-09-16'
+QUALIFY
+ dup_group_size > 1
+ORDER BY
+ DATE(time, 'America/Los_Angeles'),
+ event_name,
+ time
+ ```
+
+ ### Expanding `properties` JSON for specific events
```
argenisf: What's the goal? I think that with expanding JSON properties, the idea is to have each property be its own column, and this query would be more like each row being a separate property: 
image

argenisf (changes requested):
Mind looking at the 2 comments on the code?

argenisf (approved):

PR #2197, opened 2025-10-29
remove nesting limit

jerrylin3321:
Not enforced by Mixpanel: https://mixpanel.slack.com/archives/C022YCH9PFT/p1745335753639219?thread_ts=1745324135.065789&cid=C022YCH9PFT


jerrylin3321:
Note: the actual nesting limit (from Arb) is 10 layers. but not sure we need to add that into these docs per se

PR #2227, opened 2025-11-19
remove box shadow on dark mode toc

coltonb-mixpanel:
Removes the box shadow from this part of the table of contents when in dark mode:

| Before | After |
| - | - |
|Screenshot 2025-11-19 at 1 42 48 PM | Screenshot 2025-11-19 at 1 43 52 PM |


pages/overrides.scss
```diff
@@ -1871,6 +1871,9 @@ article {
   .nextra-toc {
     background-color: colors.$darkbg;
     // Remove all hardcoded white backgrounds and shadow filters
+    ._shadow-\[0_12px_16px_rgb\(var\(\--nextra-bg\)\)\] {
+      box-shadow: unset;
+    }
```
coltonb-mixpanel, line 1876: This is hacky but:
- This is how it's solved for [light mode](https://github.com/mixpanel/docs/blob/3887e74f0f7e0e135f642825410b718ad68ba4f3/pages/overrides.scss#L1074-L1078)
- None of the other classes on the element are particular enough to make this less fragile.

dhaval-valotia (approved):

PR #2217, opened 2025-11-12
API specs for flags endpoints

efahk:

PR #2241, opened 2025-11-24
🚩 Feature flags api spec

joshua-koehler:
Taking over from @efahk's https://github.com/mixpanel/docs/pull/2217. It had an issue where the openapi docs didn't show up. I've remedied this by linking via the reference `filename`.md and `operationId` in the openapi spec.


package.json
```diff
@@ -50,6 +50,7 @@
     "autoprefixer": "10.4.21",
     "cspell": "9.2.1",
     "postcss": "8.5.6",
+    "rdme": "^10.5.3",
```
tiffanyqi, line 53: hmm what is this package used for?

joshua-koehler: When CI failed, it indicated that we should use this readme CLI tool locally to auto-fix the reference headers.  That message is less useful when the tool doesn't exist in the project though!  So I've added it as a dev dependency.

openapi/src/feature-flags.openapi.yaml
```diff
@@ -0,0 +1,341 @@
...
+  version: 1.0.0
+servers:
+  - $ref: ./common/app-api.yaml#/server
+tags:
+  - name: Get Variant Assignments
+    description: Evaluate feature flags for a specific user
+  - name: Get Flag Definitions
+    description: Retrieve feature flag definitions
+paths:
+  /flags:
```
msiebert, line 21: I think this generates a path `api/app/flags`, which is not correct, it should be `/flags`

openapi/src/feature-flags.openapi.yaml
```diff
@@ -0,0 +1,341 @@
...
+    get:
+      operationId: get-variant-assignments
+      security:
+        - ProjectSecret: []
+      parameters:
```
msiebert, line 26: we also need the `token` param, which is their project token. We send `mp_lib` and `$lib_version` as well for our sdks, not sure if we want them to be sending something for those

msiebert (commented):
both comments apply to both endpoints

openapi/src/feature-flags.openapi.yaml
joshua-koehler: Got it, so we send both `token` and a Basic auth which encodes that same token?

openapi/src/feature-flags.openapi.yaml
joshua-koehler: Good find!  Working on this.

Also working with @tiffanyqi to see what's wrong with the readme app auth.  Worst case I'll just manually QA by copy-pasting the generated code and trying it locally, that will catch any url issues, etc.

package.json
tiffanyqi: hmmm. I wonder if it would be better just to rely on `npx rdme` rather than just installing `rdme` as a package, given we are only using it for that purpose 🤔

openapi/src/feature-flags.openapi.yaml
tiffanyqi: i haven't been able to figure out readme auth across the board historically, although it should be straightforward imo. if your app isn't working because of auth, it might not be your PR's fault lol. feel free to test it in other parts of readme before testing it on this one

openapi/src/feature-flags.openapi.yaml
msiebert: yep, we want it in the query string to make it easy to parse out in the billing flow

msiebert (approved):

package.json
joshua-koehler: To be clear, we need `rdme` installed inorder to use it via `npx rdme`, not sure if I'm missing something here, but I don't believe it was available as an npm executable prior to bringing it into the project.

openapi/src/feature-flags.openapi.yaml
joshua-koehler: I got things working 🎉

Turns out it was an incorrect base url.  Unhelpfully, we return 401's for any kind of mismatch, likely because of the ever-painful principle of _security through obscurity_.

PR #2255, opened 2025-12-05
[incr pipelines][DINO-284] enable sync by default for new schematized pipelines

howardchen11:
1. enable sync by default for new schematized pipelines
2. disable modifying sync

### DIFF
--- a/openapi/src/data-pipelines.openapi.yaml
+++ b/openapi/src/data-pipelines.openapi.yaml
@@ -343,6 +343,7 @@ components:
           type: string
           enum:
             - gcs-raw
+          default: gcs-raw
         trial:
           type: boolean
           default: false
@@ -484,6 +485,7 @@ components:
           type: string
           enum:
             - s3-raw
+          default: s3-raw
         trial:
           type: boolean
           default: false
@@ -644,6 +646,7 @@ components:
           type: string
           enum:
             - azure-raw
+          default: azure-raw
         trial:
           type: boolean
           default: false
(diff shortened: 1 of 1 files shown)


mherrman:
The changes look good but a few questions i had before we merged these changes.

1. It sounds like the intention is to have all new schematized pipelines have sync enabled correct? What does this mean for schematized date-ranged pipelines that get created? Will they also automatically have sync enabled? I know previously we didnt really want them to have it enabled but dunno if it matters as much with the incremental system.
2. Just a follow up for my knowledge, were json pipelines the same in that sync was already automatically enabled or was that not changing?

howardchen11:
> The changes look good but a few questions i had before we merged these changes.
> 
> 1. It sounds like the intention is to have all new schematized pipelines have sync enabled correct? What does this mean for schematized date-ranged pipelines that get created? Will they also automatically have sync enabled? I know previously we didnt really want them to have it enabled but dunno if it matters as much with the incremental system.
> 2. Just a follow up for my knowledge, were json pipelines the same in that sync was already automatically enabled or was that not changing?

1. `It sounds like the intention is to have all new schematized pipelines have sync enabled correct?` Yes
2. `What does this mean for schematized date-ranged pipelines that get created?` sync does not apply to date-range so it does not matter; the sync will be set to false in the codebase for date-range
3. json pipelines follows the same rule: all new recurring & non-recurring will be sync (no non-sync recurring anymore)

mherrman (approved):

mherrman:
awesome thanks! just approved

PR #2289, opened 2026-01-05
add warning for default pageview event

spjtls9:


pages/docs/tracking-methods/sdks/javascript.mdx
```diff
@@ -347,6 +347,10 @@ mixpanel.track_pageview({
     });
 ```
 
+<Callout type="warning">
+    The default page view event (`$mp_web_page_view`) cannot be merged with other events.
```
jerrylin3321, line 351: Curious, do you think we should add this here instead? https://docs.mixpanel.com/docs/data-governance/lexicon#merging-data

IlhanMP: I think it would be helpful to flag that none of the Autocapture events can be merged using `$mp_web_page_view` as an example, considering it will be the one people are most familiar with. I also agree with Jerry about the placing in the Lexicon doc.

IlhanMP: This PR can be closed. I created a new for the Lexicon doc here. https://github.com/mixpanel/docs/pull/2298

myronkaifung:
Addressed here https://github.com/mixpanel/docs/pull/2298

PR #2343, opened 2026-01-31
[DINO-381] Databricks in Json pipelines

howardchen11:
We have updated the docs based on feedback in the [bug bash.](https://www.notion.so/mxpnl/Bug-Bash-2f6e0ba92562804bb373fcf69a8a4728?source=copy_link ). Also mark this page as private beta - rolling out to a project by feature flag


jerrylin3321 (approved):

pages/docs/data-pipelines/integrations/_meta.ts
```diff
@@ -2,7 +2,7 @@ export default {
   "aws-s3": "AWS S3",
   "azure-blob-storage": "Azure Blob Storage",
   "bigquery": "BigQuery",
-  "databricks": "Databricks",
+  "databricks": "Databricks (Private Beta)",
```
smultani: Do we need the "(Private Beta)" added to the name?
We already have a callout on the homepage.

The Databricks export destination is currently in Private Beta.

I'd prefer if we remove it from name.

pages/docs/data-pipelines/integrations.mdx
```diff
@@ -13,7 +13,7 @@ These integrations allow you to send your Mixpanel data into a data warehouse.
   <Cards.Card icon title="AWS S3" href="/docs/data-pipelines/integrations/aws-s3" />
   <Cards.Card icon title="Azure Blob Storage" href="/docs/data-pipelines/integrations/azure-blob-storage" />
   <Cards.Card icon title="BigQuery" href="/docs/data-pipelines/integrations/bigquery" />
-  <Cards.Card icon title="Databricks" href="/docs/data-pipelines/integrations/databricks" />
+  <Cards.Card icon title="Databricks (Private Beta)" href="/docs/data-pipelines/integrations/databricks" />
```
smultani: Same comment as above

smultani:
Added a few comments. Looks good to me otherwise.

pages/docs/data-pipelines/integrations/_meta.ts
jerrylin3321: +1

PR #2387, opened 2026-02-23
Update feature flags documentation with variant info

russell-loube-mixpanel:
Added information about variant reassignment and feature flags.


russell-loube-mixpanel:
@efahk keep me honest here, but got some customer feedback that this wasn't clear

efahk (approved):

PR #2390, opened 2026-02-24
Update benchmarks.mdx

amymadden-bit:


kurbycchua (commented):
Just checking if there is just 1 report or potentially multiple vertical reports that we might be missing on, so holding off this approval until I get a reply here: https://mixpanel.slack.com/archives/C025PHLKDEW/p1771982301232719?thread_ts=1771950902.355969&cid=C025PHLKDEW

kurbycchua (approved):

PR #2426, opened 2026-03-05
Update notes on billing for deletes in warehouse syncs

mherrman:
Clarified billing events related to deletes in warehouse workflows.


mherrman:
removing this b/c ingestion time billing means its no longer true

Tofufu (approved):

PR #2440, opened 2026-03-09
Add mixpanel implementation skill

junhouse:
Adding the new implementation skill that AI agents can reference.


argenisf (approved):
Looks good, the only thing is that I don't see references to it within our docs, so I guess it's just available at the root like: https://docs-git-add-implementation-skill-mixpanel.vercel.app/implementation-skill.md

Thoughts on where it will be linked to?

junhouse:
> Looks good, the only thing is that I don't see references to it within our docs, so I guess it's just available at the root like: https://docs-git-add-implementation-skill-mixpanel.vercel.app/implementation-skill.md
> 
> Thoughts on where it will be linked to?

@argenisf Thanks for taking a look! We will link this skill from our in-product onboarding! But I think it will be a good idea to add this to Docs page, perhaps under the implementation section.

PR #2454, opened 2026-03-11
Fix hoverstate on button

myronkaifung:


pages/docs/reports/apps/impact.mdx
```diff
@@ -24,9 +24,31 @@ Here are some of the sample questions you can answer in Impact:
...
+<div
+  style={{
+    position: "relative",
+    paddingBottom: "64.90384615384616%",
+    height: 0,
+  }}
+>
+  <p>
+    <iframe
+      src="https://www.youtube-nocookie.com/embed/wH-dzEn02rw"
```
tiffanyqi: looks like there's an issue with the frame-src, can you do me a favor and add `https://www.youtube-nocookie.com` to `middleware.ts` after `frame-src` within the `Content-Security-Policy`? 🙏🏼

myronkaifung: This actually should have been a part of another branch. My mistake. Reverted this block of code just now.

tiffanyqi (approved):

PR #2466, opened 2026-03-17
Add test steps

junhouse:
This PR adds `test-app` folder to be used when testing mixpanel-skill. This is important to prevent any prompt drift in mixpanel-skill.

`test-app` doesn't need thorough review since it's just an example code.

Screenshot 2026-03-17 at 12 24 55 PM

### DIFF
--- a/public/mixpanel-skill/readme.md
+++ b/public/mixpanel-skill/readme.md
@@ -81,3 +81,9 @@ These phases apply to Full Implementation mode only. Quick Start uses Live View 
 - **Not legal advice.** The compliance and privacy guardrails in `SKILL.md` are implementation defaults, not legal guidance. Customer policy and counsel are the authoritative source for consent and data residency requirements.
 - **No enforcement mechanism.** The skill guides the agent to gate phases and reject shortcuts, but a customer who overrides the agent can bypass any guardrail. The skill documents the risk, not the enforcement.
 - **Developer Handoff Spec is unverified.** When no codebase access is available, the generated specification cannot be tested for correctness. The agent fills the template with session context, but cannot verify the code compiles, runs, or produces events in Live View. The developer receiving the spec must validate it.
+
+---
+
+## Testing
+
+<!-- TODO: Add testing guidance for validating changes to mixpanel-skill -->
--- a/public/mixpanel-skill/readme.md
+++ b/public/mixpanel-skill/readme.md
@@ -86,4 +86,67 @@ These phases apply to Full Implementation mode only. Quick Start uses Live View 
 
 ## Testing
 
-<!-- TODO: Add testing guidance for validating changes to mixpanel-skill -->
+The skill includes a test harness in `tests/` to validate that an AI agent can correctly implement Mixpanel tracking when given the skill.
+
+### Test App
+
+`test-app/` is a minimal React + Express e-commerce demo (Cinder & Bloom Coffee) with:
+- Product catalog, cart, checkout flow
+- Login/signup authentication flow
+- No Mixpanel pre-installed
+
+This provides a realistic codebase for the agent to implement tracking against.
+
+### Test Files
+
+| File | Purpose |
+|---|---|
+| `tests/01-implement.txt` | Prompt that instructs an agent to implement Mixpanel using the skill |
+| `tests/02-evaluate.txt` | Evaluation criteria for grading the agent's implementation |
+
+### Running a Test
+
+1. **Start a fresh agent session** with access to this repo
+2. **Feed it `01-implement.txt`** as the initial prompt
+3. **Let the agent complete** its implementation (it should follow the Quick Start flow)
+4. **Feed it `02-evaluate.txt`** to have the agent self-evaluate its work
+5. **Review the JSON output** — `pass: true` means all criteria passed
+
+### Evaluation Criteria
+
+The evaluation checks 14 criteria across 5 categories:
+
+**SDK Setup (1-3)**
+- Mixpanel initialized (SDK loaded + init() called)
+- At least one event tracked
+- Real project token used (not placeholder)
+
+**Identity Management (4-7)**
+- `identify()` called on login/signup with stable user ID
+- `reset()` called on logout
+- `identify()` called BEFORE tracking signup event
+- Email NOT used as user ID
+
+**Naming Conventions (8-9)**
+- Event names use `snake_case`
+- Property names use `snake_case`
+
+**Data Quality (10-12)**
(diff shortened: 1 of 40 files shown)


public/mixpanel-skill/readme.md
```diff
@@ -80,4 +80,73 @@ These phases apply to Full Implementation mode only. Quick Start uses Live View
...
 - **Not legal advice.** The compliance and privacy guardrails in `SKILL.md` are implementation defaults, not legal guidance. Customer policy and counsel are the authoritative source for consent and data residency requirements.
 - **No enforcement mechanism.** The skill guides the agent to gate phases and reject shortcuts, but a customer who overrides the agent can bypass any guardrail. The skill documents the risk, not the enforcement.
-- **Developer Handoff Spec is unverified.** When no codebase access is available, the generated specification cannot be tested for correctness. The agent fills the template with session context, but cannot verify the code compiles, runs, or produces events in Live View. The developer receiving the spec must validate it.
\ No newline at end of file
+- **Developer Handoff Spec is unverified.** When no codebase access is available, the generated specification cannot be tested for correctness. The agent fills the template with session context, but cannot verify the code compiles, runs, or produces events in Live View. The developer receiving the spec must validate it.
+
+---
+
+## Testing
```
junhouse, line 87: How to test/evaluate.

public/mixpanel-skill/tests/01-implement.txt
```diff
@@ -0,0 +1,8 @@
+Set up Mixpanel tracking in this project: `public/mixpanel-skill/test-app`. My project token is mixpanel-token. 
+Use the Mixpanel setup skill: public/mixpanel-skill/skill.md
```
tiffanyqi, line 2: the file is readme.md right?

public/mixpanel-skill/tests/02-evaluate.txt
```diff
@@ -0,0 +1,40 @@
+Role: You are a strict QA evaluator testing an AI coding agent. 
+
+Task: Evaluate the agent's code changes in public/mixpanel-skill/test-app against the following criteria.
+
+Criteria:
+
+## SDK Setup
+1. Did the agent successfully initialize Mixpanel (SDK loaded + init() called)?
+2. Did the agent track at least one event?
+3. Did the agent use a real project token (not a placeholder like 'YOUR_PROJECT_TOKEN')?
```
tiffanyqi, line 10: or maybe the one that you provided?

public/mixpanel-skill/tests/02-evaluate.txt
```diff
@@ -0,0 +1,40 @@
...
+
+Output:
+Output ONLY a raw JSON object with this exact schema. Do not use markdown formatting or backticks around the JSON.
+{
+  "pass": boolean,
+  "failed_criteria": ["List the criteria numbers that failed. Leave empty if pass is true."],
+  "reason": "A strict 1-2 sentence explanation of the ruling."
+}
+
+Once finished, revert all changes made to public/mixpanel-skill/test-app`.
```
tiffanyqi, line 40: can you clarify this one? is the plan for this to ensure the tests pass, and then it will revert everything? that works for the most part, i'm just trying to think through test failures and reproduction steps.

tiffanyqi: for testing, will it also run the app, or just check if there is code there?

public/mixpanel-skill/test-app/server/package.json
```diff
@@ -0,0 +1,19 @@
...
+  "main": "src/index.js",
+  "scripts": {
+    "dev": "nodemon src/index.js",
+    "start": "node src/index.js"
+  },
+  "dependencies": {
+    "cors": "^2.8.5",
+    "dotenv": "^16.4.5",
+    "express": "^4.19.2",
+    "morgan": "^1.10.0"
```
tiffanyqi, line 14: this is mostly fine to me, however if we are going to use a test app does it make sense to keep these dependencies up to date? or can we turn off all alerting for it?

public/mixpanel-skill/test-app/server/src/data/products.js
```diff
@@ -0,0 +1,54 @@
...
+  {
+    id: "ethiopia-yirgacheffe",
+    name: "Ethiopia Yirgacheffe",
+    origin: "Ethiopia",
+    roast: "Light",
+    tastingNotes: ["Jasmine", "Lemon", "Honey"],
+    priceCents: 1899,
+    size: "12 oz",
+    grindOptions: ["Whole Bean", "Espresso", "Drip", "French Press"],
+    image: "/images/ethiopia.jpg",
```
tiffanyqi, line 11: will there be an issue if this image doesn't render?

tiffanyqi (commented):
left some questions!

public/mixpanel-skill/tests/01-implement.txt
junhouse: no, the skill is `skill.md` file.

tiffanyqi: gotcha, i see that there's a skill.md file already

public/mixpanel-skill/tests/02-evaluate.txt
junhouse: Basically, `test-app` contains sample code the agent can use to implement tracking. Once it's done, we will evaluate the code changes - verifying tracking code an agent added are correct. Once that's done, we have to revert the changes so we can use the same `code` for evaluation purposes in future.

junhouse: No, we don't have to run the app since we are mostly evaluating code added.

public/mixpanel-skill/test-app/server/src/data/products.js
junhouse: no, because we are not interested in app actually running. We are only evaluating code added by an agent.

public/mixpanel-skill/test-app/server/package.json
junhouse: Actually these dependencies don't matter, maybe I should rid of it.

public/mixpanel-skill/tests/02-evaluate.txt
tiffanyqi: in that case, do we need any of the server code? since the plan isn't to actually run? (unless we wanted tracking in the server code as well)

junhouse: Yes, server code for server side tracking!

PR #2537, opened 2026-04-02
Add Custom Roles documentation

dongjae93:
## Summary
- Adds a new "Custom Roles" section to the Roles & Permissions page documenting the Enterprise-plan feature
- Covers creating, editing, cloning, and deleting custom roles
- Includes a permission categories table with all 6 configurable categories
- Adds callouts for Enterprise plan requirement, base permissions, cloning constraints, and deletion behavior

## Test plan
- [ ] Verify the Vercel preview renders the new section correctly
- [ ] Check that anchor links (e.g., `#custom-roles`, `#permission-categories`) work
- [ ] Confirm tables and callouts display properly
- [ ] Review content accuracy with the custom roles feature team

🤖 Generated with [Claude Code](https://claude.com/claude-code)


pages/docs/orgs-and-projects/roles-and-permissions.mdx
caracheng, line 1: > All custom roles automatically include base permissions: viewing reports, editing visibility of entities you created, and locking/unlocking entities you created. 

The terminology of "locking/unlocking" entities isn't familiar language that we use in the docs. Is this a reference to sharing?

dongjae93: ah yes. That's what the permissions is called so I called it that but it is for sharing

caracheng: > Once created, the custom role will appear alongside default roles when assigning project roles to users.

Can we explicitly mention that custom roles can be applied to teams as well?

dongjae93: actually it cannot. Custom roles is only at project level grant team is at org level. Teams can only have org grant roles. This is a good callout though I'll add to explain that custom roles are just for project level grants

pages/docs/orgs-and-projects/roles-and-permissions.mdx
```diff
@@ -299,3 +299,87 @@ To delete a team in Mixpanel:
...
 
 ![image](/Del_Team.png "Delete Team")
+
+## Custom Roles
+
+<Callout type="info">
+Custom Roles is available to organizations on **Enterprise** plans. See our [pricing page](https://mixpanel.com/pricing/) for more details.
+</Callout>
+
+Custom roles let you create specialized access profiles beyond the four default project roles (Owner, Admin, Analyst, Consumer). With custom roles, you can define granular permissions tailored to specific job functions or workflows. Each project can have up to **5 custom roles**.
```
kurbycchua, line 318: I thought we are renaming Analyst? Or maybe Not yet? just checking to make sure we account for that

pages/docs/orgs-and-projects/roles-and-permissions.mdx
```diff
@@ -299,3 +299,87 @@ To delete a team in Mixpanel:
...
+
+Only project **Owners** and **Admins** can create and manage custom roles.
+
+### Creating a Custom Role
+
+To create a custom role:
```
kurbycchua, line 324: Wondering if a screenshot would help here?

pages/docs/orgs-and-projects/roles-and-permissions.mdx
kurbycchua: 1. Also confused by the "locking/unlocking" - if it's sharing entities we should just call it sharing
2. If custom roles can't work for teams - this might be confusing? How do I setup teams + custom roles at project level for groups of users? Do I...

Create Org Level Teams - assign user to these teams?
Create Custom Project Roles - and assign user to the roles?

Would that be sort of a double effort?

allen-wang-Mixpanel: 1. The existing permissions system before DJ and I started on this project already had the naming like `edit_locked_created_entity` and `edit_locked_entity`, however if this is a point of confusion it should be small effort to rename it
2. Custom roles is per project - that being said, you can go to Org Settings/Teams page to edit the team's role for each project, possibly giving the team a custom role there. I'd imagine yes if you want to give a group of users a custom role grouping them into a team is the easiest.

pages/docs/orgs-and-projects/roles-and-permissions.mdx
allen-wang-Mixpanel: I think the idea is to revamp to `Contributor`, however I do see various references which still use `Analyst` which is definitely a point of confusion, there will have to be broader effort to maintain a consistency across our systems.

allen-wang-Mixpanel: but for this for now, since the custom roles tab says contributor, let's use contributor here as well

pages/docs/orgs-and-projects/roles-and-permissions.mdx
allen-wang-Mixpanel: might be helpful to just link Sharan's demo: https://www.loom.com/share/b206a17409c6487f9e5b911eb0f8ab59

allen-wang-Mixpanel (commented):
Left some comments

pages/docs/orgs-and-projects/roles-and-permissions.mdx
dongjae93: we went back to using analyst 👍

allen-wang-Mixpanel (approved):
LGTM!

PR #2553, opened 2026-04-06
Update audit log date format in changelog

mherrman:


myronkaifung:
Timezone issue addressed in https://github.com/mixpanel/docs/pull/2554

PR #2611, opened 2026-04-20
[SRFE-9055] Note that distinct id is not required

jerrylin3321:
Jira: https://mixpanel.atlassian.net/browse/SRFE-9055

### DIFF
--- a/reference/Ingestion API/events/import-events.md
+++ b/reference/Ingestion API/events/import-events.md
@@ -85,6 +85,8 @@ If the time value is set in the future, it will be overwritten with the current 
 
 If the event is not associated with any user, set distinct\_id to the empty string. Events with an empty distinct\_id will be excluded from all behavioral analysis.
 
+If no distinct\_id is provided, Mixpanel will use the ip address of the incoming request and compute a distinct\_id using a hash function.
+
 To prevent accidental implementation mistakes, we disallow the following values for distinct\_id:
 
 ```


chris-bonadonna-mixpanel (approved):

santigracia (changes requested):
@jerrylin3321 thanks for the update! couple of questions:

1- There is a required red label next to distinct_id in the body params in this page:
https://developer.mixpanel.com/reference/import-events when you click on "Add object".

Image
Is this correct?

This required is coming from here: https://github.com/mixpanel/docs/blob/main/openapi/src/ingestion.openapi.yaml#L104

Readme generates the Try Me! examples with our openapi spec in yaml

2- This page https://developer.mixpanel.com/reference/event-deduplication 
is showing required for distinct_id. Is this correct?

Image

jerrylin3321:
> @jerrylin3321 thanks for the update! couple of questions:
> 
> 1- There is a required red label next to distinct_id in the body params in this page: https://developer.mixpanel.com/reference/import-events when you click on "Add object".
> 
> Image Is this correct?
> This required is coming from here: https://github.com/mixpanel/docs/blob/main/openapi/src/ingestion.openapi.yaml#L104
> 
> Readme generates the Try Me! examples with our openapi spec in yaml
> 
> 2- This page https://developer.mixpanel.com/reference/event-deduplication is showing required for distinct_id. Is this correct?
> 
> Image

@santigracia thanks for checking this over. after thinking about it, maybe it's better to keep it as "required." It seems that in many places throughout our docs, we say that it is required. I changed this PR's scope to just add a little disclaimer noting that we'll automatically generate a distinct id if none is specified

santigracia (approved):
LGTM!

PR #2612, opened 2026-04-21
Clarify warehouse connector billing for legacy event timestamp accounts

argenisf:
Updates and deletes are billed at ingestion time for all accounts, including those on legacy event timestamp billing. The previous wording implied that all operations follow event timestamp billing for legacy accounts, which was misleading.


santigracia (approved):
Approved!

PR #2629, opened 2026-04-24
Clarify experiment visibility and implementation docs

kaan-barmore-genc-mixpanel:
## Summary
- Update the Step 1 note to explain that both Mixpanel Feature Flag experiments and exposure-event experiments populate the dropdown, and describe when each is hidden (archived flag; no exposure event in 30 days).
- Document how to hard-code an experiment name with the `$experiment:Experiment name` filter/breakdown.
- Add an intro to the Implementation for Experimentation section noting that Mixpanel Feature Flags handle exposure event tracking automatically, and frame the existing format as the path for 3rd party / in-house feature flag systems.

## Test plan
- [ ] Visually confirm the rendered Experiments page reads cleanly at the Step 1 callout and the Implementation for Experimentation section


myronkaifung (approved):

santigracia (approved):
LGTM

PR #2676, opened 2026-05-12
Docs: page views captured for heatmap are exempt from event/MTU billing

djmixpanel:
Document that $mp_web_page_view events emitted with $captured_for_heatmap: true (while Session Replay is recording) are excluded from event and MTU billing, matching the new HEATMAP_FREE_EVENT_NAMES behavior in analytics#94007. Clarifies that URL-based Conditional Rules for sampling do not add to billable event volume.


lajohn4747 (approved):

santigracia (approved):
Approved!

PR #2695, opened 2026-05-14
[DF-184] Adding Blocked and Filtered Status updates to docs

phuang-mixpanel:
Added a new subsection to drop filters about blocked and filtered status.


pages/docs/data-governance/data-clean-up.mdx
```diff
@@ -236,6 +236,12 @@ To stop dropping events, locate the filter in **Lexicon > Data Governance > Data
 
 ![image](/drop-filter-3.png)
 
+### Dropping a full event
+
+To drop an entire event, locate the event in Lexicon under the **Events** tab. Select the event and under ***Status*** toggle on ***Blocked***. Fully dropped events have a status of ***Blocked*** on lexicon while events dropped with drop filters have a status of ***Filtered***
+
+![image](/drop-filter-4.png)
```
Bruna-Liborio: I think we can remove this image here. Images can get stale and since this is just a renaming I don't think the UI reference is necessary

Bruna-Liborio (changes requested):
I think we need to update language in a few other places from dropped to blocked:
* https://docs.mixpanel.com/docs/data-governance/data-clean-up#dropping-events-that-are-not-queried
* https://docs-lhkdzjx91-mixpanel.vercel.app/docs/data-governance/lexicon#dropping-data (there is a dropping data section that needs to be updated here and we should update other mentions of drop in this doc page to be block)

pages/docs/data-governance/data-clean-up.mdx
Bruna-Liborio, line 118: We should probably update the rec here to be "Blocking and Hiding Data" and I think the link might need to be updated.

Bruna-Liborio (approved):
Nice! Found one place where we might need an update, besides that looks like everything is covered. Thanks for the image updates too! those can be annoying to keep up to date

PR #2696, opened 2026-05-14
[SDK-61] Update Android SDK EU and India Data Residency docs

santigracia:
## Summary
- MixpanelOptions is now the primary recommended approach with a code example showing serverURL set during initialization
- Added a Callout matching the Swift docs style: "Setting serverURL during initialization ensures all requests (events, profiles, and feature flags) use the correct endpoint from the start."
- AndroidManifest.xml kept as an alternative, with DecideEndpoint replaced by FlagsEndpoint
- Removed setServerURL() runtime method, no longer shown as an option in docs (same as Swift docs).


argenisf (approved):
LGTM

PR #2703, opened 2026-05-15
Flutter Session Replay beta support

tylerjroach:
Will remove "draft" PR label once we are ready to move to public beta.

Can I get confirmation that the images are in an appropriate location?


myronkaifung (approved):

argenisf (approved):
good to go!

PR #2704, opened 2026-05-15
[audit-log] new event type documentation

coltonb-mixpanel:
**Note: Final list and date of audit log types for release is not completely finalized.**

Adds documentation for all new to-be-released audit log types.

Rather than a precise UTC timestamp for event release, I've updated with "available since" to the first full date with global coverage for readability & simplicity. e.g., if we release at May 18th at 12:00 PM PDT, a full day of audit logs is available globally starting May 20th.

### DIFF
--- a/pages/docs/access-security/audit-log.mdx
+++ b/pages/docs/access-security/audit-log.mdx
@@ -33,211 +33,211 @@ Mixpanel tracks the following events in your Audit Log. The "Available Since" co
 
 ### Alerts
 
-| Event Type        | Display Name    | Description                               | Available Since      |
-| ----------------- | --------------- | ----------------------------------------- | -------------------- |
-| `alert.created`   | Alert Created   | A new alert was created on a report       | 2026-01-31T00:00:00Z |
-| `alert.deleted`   | Alert Deleted   | An alert was removed                      | 2026-01-31T00:00:00Z |
-| `alert.triggered` | Alert Triggered | An alert condition was met and triggered  | 2026-02-05T21:00:00Z |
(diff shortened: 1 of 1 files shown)


mherrman (approved):

pages/docs/access-security/audit-log.mdx
```diff
@@ -25,121 +27,219 @@ Some events are tracked only at the organization level, while most are available
...
-- **Cohorts** - 2026-04-03T19:00:00Z
-- **Data Pipelines** - 2026-04-03T19:00:00Z
-- **Warehouse Sources** - 2026-04-03T19:00:00Z
-- **Data Management** - 2026-04-03T19:00:00Z
-- **Data Exports** - 2026-04-03T19:00:00Z
-- **Service Accounts** - 2026-04-03T19:00:00Z
-- **User Authentication** - 2026-04-03T19:00:00Z
-
-Mixpanel tracks the following events in your Audit Log:
+The current audit logging contains only actions made within the Mixpanel UI. Each line in the UI will contain information on the Date and Time, User, Action, IP and Project or Org if available. Clicking on each log item you'll be able to see a raw view with additional information.
```
mherrman, line 30: i think the part about this being UI actions isnt technically true irrc right? like MCP actions are also captured if they do the thing? Initially i think this was suppose to be documenting that API actions were not captured but we added /export API so some of those are now in the logs as well.

```suggestion
Each audit log line in the UI will contain information on the Date and Time, User, Action, IP and Project or Org if available. Clicking on each log item you'll be able to see a raw view with additional information.
```

coltonb-mixpanel: ah shoot -- good point.

PR #2710, opened 2026-05-18
Update flags docs for dynamic config support in experiments

efahk:
Feature flags now support JSON config as feature flag values in experiment flags. This change updates the public documentation for consistency.


pages/docs/experiments.mdx
```diff
@@ -73,6 +73,16 @@ Mixpanel has set default automatic configurations, seen below. If required, plea
...
 
+### Variant Values
+
+Each variant can optionally carry a **value** in addition to its name, either a **string** or a **JSON** object. The variant name is what your code switches on to render the right experience. The value is configuration data the SDK returns alongside the assignment (e.g., a CTA label, a feature config object).
+
+- The value type is chosen once per flag. Every variant in the flag must use the same type, either all string or all JSON.
+- JSON values are validated when you save the flag; values that don't parse are rejected.
+- Each variant's value has a 32 KB size limit.
+
+This is the same mechanism used by [Dynamic Config](/docs/featureflags#types-of-feature-flags) flags, now available on Experiment flags as well.
```
russell-loube-mixpanel:
````suggestion
### Experiment JSON Parameters

Each variant can optionally carry a **value** in addition to its name — either a **string** or a **JSON object**. The variant name is what your code switches on to render the right experience. The value is configuration data the SDK returns alongside the assignment, so you can remotely control parameters of your app or site (copy, colors, thresholds, layout configs, pricing, etc.) without a code deploy.

For example, a variant named `treatment` might carry a JSON object value like:

```json
{ "cta_text": "Buy now", "discount": 20, "layout": "compact" }
```

Your client reads `variant.value` after assignment and uses it to drive the UI.

- The value type is set per flag. Every variant in the flag must use the same type — either all string or all JSON. You can change a flag's value type later from the flag's settings.
- JSON values are validated when you save the flag. If a value doesn't parse as valid JSON, the flag can't be saved — whether you're editing in the UI or updating it via the API.
- Each variant's value has a 32 KB size limit.

#### Example: reading a variant value, with a provided fallback value.(JavaScript SDK)

```javascript
mixpanel.flags.get_variant_value("checkout_cta_experiment", { "cta_text": "fallback_cta_text", "discount": 0, "layout": "wide" })
  .then((value) => {
    // value is the JSON object configured on the assigned variant
    document.querySelector("#cta").innerText = value.cta_text;
  });
```

This is the same mechanism used by [Dynamic Config](/docs/featureflags#types-of-feature-flags) flags, now available on Experiment flags as well.
````

File: pages/docs/featureflags.mdx [hunk 2]
```diff
@@ -46,13 +46,13 @@ Users will not be reassigned different variants when purely increasing overall r
 We support the following types of Flags:
 
 1. **Feature Gate** : Toggle a feature on or off for targeted users or all users. Useful for phased or controlled rollout.
-2. [**Experiment**](/docs/experiments) : Deliver different variant experiences (e.g., layouts, flows, pricing) to a targeted group of users. Enables measuring and analyzing impact.
+2. [**Experiment**](/docs/experiments) : Deliver different variant experiences (e.g., layouts, flows, pricing) to a targeted group of users. Enables measuring and analyzing impact. Each variant can optionally carry a **string** or **JSON** value, letting you ship per-variant configuration (e.g., a CTA label, a layout config blob) alongside the variant name.
```
russell-loube-mixpanel:
```suggestion
2. [**Experiment**](/docs/experiments) : Deliver different variant experiences (e.g., layouts, flows, pricing) to a targeted group of users. Enables measuring and analyzing impact. Each variant can optionally carry a **string** or **JSON object** value, letting you ship per-variant configuration (e.g., a CTA label, a layout config blob) alongside the variant name. See [Experiment JSON Parameters](/docs/experiments#experiment-json-parameters).
```

pages/docs/featureflags.mdx
```diff
@@ -46,13 +46,13 @@ Users will not be reassigned different variants when purely increasing overall r
...
 3. **Dynamic Config** : Configure features with flexible key-value pairs instead of just on/off. It lets you:
-    - Pass JSON payloads (e.g., `{"cta_text": "Buy now", "discount": 20}`) to customize behavior or UI dynamically
+    - Pass **string** or **JSON** payloads (e.g., `{"cta_text": "Buy now", "discount": 20}`) to customize behavior or UI dynamically
```
russell-loube-mixpanel:
```suggestion
    - Pass **string** or **JSON object** payloads (e.g., `"Buy now"` or `{"cta_text": "Buy now", "discount": 20}`) to customize behavior or UI dynamically
```

russell-loube-mixpanel (commented):
Nice work — this is close. Three small suggestions inline:

1. Rename the section to **Experiment JSON Parameters** (the original name risked overlap with our marketing language).
2. Drop the string-only example and lean into the JSON object case, which is the more compelling story.
3. Clarify what "rejected" means — i.e., the flag can't be saved via UI or API.

Also added a small JS SDK snippet so readers can see the shape of the payload end-to-end.

**One thing to confirm before merging:**
- The JS SDK method I used (`mixpanel.flags.get_variant_value`) is a placeholder — please swap in the real method name from the JS SDK reference.
- Does the value type really stay changeable after creation, or only before any assignments exist? The wording "you can change a flag's value type later" should match real behavior.

pages/docs/featureflags.mdx
```diff
@@ -46,13 +46,13 @@ Users will not be reassigned different variants when purely increasing overall r
...
     - Update values instantly without redeploying code
 
     <Callout type="info">
-        The JSON payload for Dynamic Config has a size limit of 32 KB.
+        Variant values (used by both Experiment and Dynamic Config flags) have a size limit of 32 KB per variant. Within a single flag, all variants must use the same value type, either all string or all JSON.
```
russell-loube-mixpanel:
```suggestion
        Variant values (used by both Experiment and Dynamic Config flags) have a size limit of 32 KB per variant. Within a single flag, all variants must use the same value type — either all string or all JSON object and you can switch the type later from the flag's settings.
```

russell-loube-mixpanel (approved):
lgtm maestro

PR #2725, opened 2026-05-25
Update mcp.mdx

gonsolvis-mxp:
In our docs, these are called Boards, so adding Boards to not confuse the customer as some users think dashboards are reports


santigracia:
@smultani reviewing this PR from @gonsolvis-mxp , we need clarification on the naming of Boards for MCP. In general, we always used "Boards" across all the product but the MCP is using "Dashboards".

@myronkaifung there are 2 workflows requiring approval to run

myronkaifung (approved):

PR #2729, opened 2026-05-27
docs: add Benjamini-Hochberg correction and configurable CUPED lookback

russell-loube-mixpanel:
## Summary

Updates the **Advanced Statistical Methods** section of the Experiments docs to cover two features that are now exposed in the product UI:

- **Benjamini-Hochberg correction** — a new option in the Multiple Testing Correction dropdown, alongside the existing Bonferroni. Docs explain FDR vs. FWER, the rank-based procedure, a worked 5-metric example, when to prefer BH over Bonferroni, and the constraint that only one correction method can be applied at a time. Content drawn from Kaan's internal [Benjamini-Hochberg writeup](https://www.notion.so/mxpnl/Benjamini-Hochberg-2dfe0ba9256280c2bed2c35059a58d37).
- **Configurable CUPED Pre-Exposure Period** — previously the docs said "a pre-exposure period of your choosing" without enumerating options. Now lists the five options (1 Week, 2 Weeks default, 4 Weeks, 60 Days, 90 Days) and gives guidance on picking a window.

Also adds a new row in the "Methods at a Glance" table for Benjamini-Hochberg and a small clarifying tail to the Bonferroni row's "When to Use" cell to contrast the two methods.

No other sections were touched.

## Test plan

- [ ] Render preview on Vercel and confirm the new BH section, example table, and CUPED paragraph render correctly
- [ ] Kaan to verify BH explanation matches the implementation
- [ ] Confirm the listed CUPED lookback options match what ships in the UI dropdown

🤖 Generated with [Claude Code](https://claude.com/claude-code)


pages/docs/experiments.mdx
```diff
@@ -309,7 +310,37 @@ Bonferroni Correction addresses this by making the significance threshold strict
 - You have multiple treatment variants competing against control
 - You want higher confidence that significant results are real
 
-Bonferroni Correction is conservative. It reduces false positives but also makes it harder to detect true effects. If you have a single metric that matters most to you, you may prefer to focus on that primary metric without correction.
+Bonferroni Correction is conservative. It reduces false positives but also makes it harder to detect true effects. If you have a single metric that matters most to you, you may prefer to focus on that primary metric without correction. If you're testing many metrics and Bonferroni feels too strict, consider Benjamini-Hochberg below.
+
+### Benjamini-Hochberg Correction
+
+Benjamini-Hochberg (BH) is a more balanced alternative to Bonferroni for handling multiple comparisons. Instead of controlling the probability that *any* significant result is a false positive (the family-wise error rate that Bonferroni targets), BH controls the **false discovery rate (FDR)** — the expected *proportion* of your "winners" that are actually false positives.
```
kaan-barmore-genc-mixpanel: We never say "FDR" again, so probably don't need to include the abbreviation here.

pages/docs/experiments.mdx
```diff
@@ -363,6 +394,8 @@ The core insight is that users who had high engagement or revenue *before* the e
 
 **How it works:** For each user, Mixpanel looks at their metric value during a pre-exposure period of your choosing and their metric value during the experiment. If these values are strongly correlated (users with high pre-experiment values tend to have high post-experiment values), CUPED uses this relationship to reduce variance in the experiment results. The mean values remain unchanged—CUPED only tightens the confidence intervals. This is applied to all metric categories: primary, secondary, and guardrail metrics.
 
+**Configuring the pre-exposure period:** When you enable CUPED, you can choose the lookback window under **Configuration → CUPED Pre-Exposure Period**. The available options are **1 Week**, **2 Weeks** (default), **4 Weeks**, **60 Days**, and **90 Days**. Longer windows give CUPED more historical data per user, which can improve variance reduction when behavior is stable over time, but they also exclude users whose history doesn't reach that far back. Shorter windows include more users but may capture less predictive signal. Two weeks is a good default for most experiments; consider a longer window if your metric has slow-moving or seasonal patterns.
```
kaan-barmore-genc-mixpanel: This isn't correct. When you set the pre-exposure period, that's the length of time we'll query going back. The more you expand this pre-exposure period, the more users we'll actually pick up. We only filter out people who were active during this period but weren't exposed to the experiment.

I think the right point to make here is that you want to choose the pre-exposure period so that it doesn't overlap with other experiments on the same user population as this one, otherwise there is loss of sensitivity and potential bias. And long enough to capture user behavior (say, if some of your users are active only once a month, you want to make sure you choose a range long enough to see them).

I also asked Claude for feedback, I'll add a relevant point it gave:

> A longer window is not always better — stale behavior from far in the past can be less predictive of current behavior, and flooding the population with X = 0 (new users, infrequent users) drags ρ toward zero.

pages/docs/experiments.mdx
russell-loube-mixpanel: Removed — good catch.

pages/docs/experiments.mdx
russell-loube-mixpanel: Good correction — I had it backwards. Updated to: longer windows query further back and pick up *more* users; guidance now covers avoiding overlap with other experiments, ensuring infrequent users are captured, and the stale-data caveat you mentioned (including the ρ-toward-zero point from Claude's feedback).

russell-loube-mixpanel: (generated by claude)

kaan-barmore-genc-mixpanel (approved):

myronkaifung (approved):

